PR #4, opened 2024-10-05
Review - Improve changes and move to rag_agent

chimosky:
Move RAG logic to a new file, use a class
for the functionalities.

Combine requirements.txt, no need for two separate files.

**TODO**
- [x] Migrate from RetrivevalQA as it's [deprecated](https://python.langchain.com/v0.2/docs/versions/migrating_chains/retrieval_qa/).
- [x] Make sure langchain tools used are up to date.
- [x] Create docs folder and add docs for RAG there.
- [x] `self.prompt` seems to be unused, why? I'm guessing migrating from RetrievalQA might solve this.
- [x] Redirect output to right sources instead of using print statements.

@kshitijdshah99 @XXXJumpingFrogXXX kindly review.
I'll leave the other changes to you Kshitij.

This PR is open and we'll use it to track the **TODO** above, and will merge once you've made the necessary changes.

PR #6, opened 2024-10-09
Migrated from RetrievalQA, removed print statements

kshitijdshah99:
I have added paths of all the docs and modified the code as per our requirements. This Fixes #4

### DIFF
--- a/rag_agent.py
+++ b/rag_agent.py
@@ -1,66 +1,61 @@
 # Pippy's AI-Coding Assistant
-# Date: 2023-02-15
 # Uses Llama3.1 model from Ollama
 
 import os
-import warnings, logging
-
+import warnings
 from langchain_community.vectorstores import FAISS
-from langchain_community.embeddings import HuggingFaceEmbeddings, OpenAIEmbeddings
+from langchain_community.embeddings import HuggingFaceEmbeddings
 from langchain_community.document_loaders import PyMuPDFLoader, TextLoader
-from langchain.chains import RetrievalQA
-from langchain_core.prompts import ChatPromptTemplate
+from langchain_core.runnables import RunnablePassthrough
+from langchain_core.output_parsers import StrOutputParser
 from langchain_ollama.llms import OllamaLLM
-
+from langchain.prompts import ChatPromptTemplate
 
 # Suppress future warnings related to deprecations
 warnings.filterwarnings("ignore", category=FutureWarning)
 warnings.filterwarnings("ignore", category=DeprecationWarning)
 
+# Document paths
 document_paths = [
-    './docs/Pygame Documentation.pdf'),
-    './docs/Python GTK+3 Documentation.pdf'),
+    '/home/kshitij/Downloads/Sugarlabs/Pippy-Activity/Pygame Documentation.pdf',
+    '/home/kshitij/Downloads/AI-model/Python GTK+3 Documentation.pdf',
 ]
 
-PROMPT = """
-You are a highly intelligent Python coding assistant with
-access to both general knowledge and specific Pygame documentation.
-1. You only have to answer Python and GTK based coding queries.
-2. Prioritize answers based on the documentation when the query is related to it.
-   However make sure you are not biased towards documentation provided to you.
-3. Make sure that you don't mention words like context or documentation
-   stating what has been provided to you.
-4. Provide step-by-step explanations wherever applicable.
-5. If the documentation does not contain relevant information, use your
-   general knowledge.
-6. Always be clear, concise, and provide examples where necessary.
-"""
+# Revised Prompt Template to avoid mentioning the source
+PROMPT_TEMPLATE = """
+You are a highly intelligent Python coding assistant. 
+You are ONLY allowed to answer Python and GTK-based coding questions. 
+1. Focus on coding-related problems, errors, and explanations.
+2. Use the knowledge from the provided Pygame and GTK documentation without explicitly mentioning the documents as the source.
+3. Provide step-by-step explanations wherever applicable.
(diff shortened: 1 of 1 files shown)


rag_agent.py
chimosky, line 1: What's the purpose of this commit?

rag_agent.py
```diff
@@ -71,74 +66,72 @@ def setup_vectorstore(self, file_paths):
...
+            'input', 'string', 'number', 'list', 'tuple', 'dictionary', 'module', 
+            'import', 'turtle', 'indentation', 'syntax', 'Pygame', 'math', 'shape', 
+            'for', 'while', 'repeat', 'class', 'object', 'method', 'attribute', 
+            'operator', 'condition', 'true', 'false', 'boolean', 'integer', 'float', 
+            'type', 'index', 'range', 'len', 'append', 'pop', 'remove', 'sort', 
+            'reverse', 'random', 'sleep', 'time', 'pygame', 'GTK', 'logic', 'numbers', 'correct', 'answer'
+        ]
+        return any(keyword.lower() in question.lower() for keyword in coding_keywords)
+
+
```
chimosky, line 103: We didn't agree on this change, why is it here?

rag_agent.py
```diff
@@ -71,74 +66,72 @@ def setup_vectorstore(self, file_paths):
...
+            | self.prompt  
+            | self.model   
+            | StrOutputParser()  
+        )
+
         while True:
-            # We'll need to change this as time goes on as we won't
-            # get user input this way.
-            question = input("Enter your question (or type 'exit' to quit): ").strip()
+            question = input("Hey there, let me know your coding doubt").strip()
```
chimosky, line 123: Doesn't seem like a great prompt.

rag_agent.py
kshitijdshah99: Sorry that was a typo I'll fix it

rag_agent.py
kshitijdshah99: Ok then I'll remove this out from our code.

rag_agent.py
kshitijdshah99: Actually the prompt would be empty as the child would be adding his query in the text window and then we'll pass it to our model through this, I had added it just for reference.
I can keep it empty or write something like
```What Python magic are we working on today?```
```Let’s code something cool! Ask me!```
Let me know what are your opinions on this

Comment (chimosky) [reply rag_agent.py:123]: If it's meant to be empty, then leave it empty.

File: rag_agent.py
Comment (chimosky) [reply rag_agent.py:1]: Doesn't look like a typo, I'm referring to the change in d926173.

File: rag_agent.py [hunk 3]
```diff
@@ -1,66 +1,61 @@
...
-class RAG_Agent():
+class RAG_Agent:
     def __init__(self):
         self.model = None
         self.retriever = None
-        self.prompt = ChatPromptTemplate.from_template(TEMPLATE)
+        self.prompt = ChatPromptTemplate.from_template(PROMPT_TEMPLATE)
 
-    def set_model(self, model):
+    def set_model(self, model="llama3.1"):
```
chimosky, line 43: The reason why this function exists is to set self.model to whatever model is passed, a default value for the model can be set in `init`, not here.
This is a setter function so a model kwarg defeats the purpose.

rag_agent.py
```diff
@@ -1,66 +1,61 @@
...
+        """Set the Llama 3.1 model from Ollama."""
```
chimosky, line 44: My comment above makes this redundant, you can set a docstring that's better as you now understand why the function exists.

rag_agent.py
kshitijdshah99: Yeah true

chimosky:
You've set an initial value for `self.model` in a640e38, but haven't removed the kwarg you passed to `set_model`.

Your initial value means that `set_model` doesn't need to be called if you're setting it to the same value as the one that's been set.

chimosky:
Just looked at c45998c, I didn't say you needed to remove the function, it's not redundant. 

I said **"You've set an initial value for self.model in https://github.com/sugarlabs/sugar-ai/commit/a640e38fdfc421d4f38be663bf85936585335db5, but haven't removed the kwarg you passed to set_model."**

Please let me know if you don't understand what I meant.

kshitijdshah99:
ok so you want me to do this changes?
```
def __init__(self, model="llama3.1"):
        """
        Initializing the RAG agent with a default model like Llama3.1.
        """
        self.model = OllamaLLM(model=model)
        self.retriever = None
        self.prompt = ChatPromptTemplate.from_template(PROMPT_TEMPLATE)

    def set_model(self, model):
        self.model = OllamaLLM(model=model)
        ```

**Comment by chimosky** [comment]
Yes, that's what I want.

The init docstring is stating the obvious, if that's all it does then I don't think we need it.

File: rag_agent.py [hunk 5]
```diff
@@ -43,6 +43,10 @@ def __init__(self, model="llama3.1"):
         self.retriever = None
         self.prompt = ChatPromptTemplate.from_template(PROMPT_TEMPLATE)
 
+    def set_model(self, model):
+        """To Set a new model if needed."""
```
chimosky, line 47: This is redundant, the function is self explanatory.

chimosky (changes requested):

rag_agent.py
chimosky: I also said this **The init docstring is stating the obvious, if that's all it does then I don't think we need it.**, seems you didn't notice it.

chimosky:
Fixed the issues with redundant comments in 3ca1ff7.

PR #7, opened 2024-10-10
FastAPI Framework Setup

XXXJumpingFrogXXX:
This PR includes the following changes:
(1) Created the latest main.py file and completed some basic FastAPI settings in it.
(2) Renamed the original main.py file to original_main.py.
(3) Kept the existing piggy directory and created a chat directory to establish separate routers and APIs for each project.

### DIFF
--- a/chat/router.py
+++ b/chat/router.py
@@ -1,0 +1,110 @@
+from fastapi import APIRouter
+from pydantic import BaseModel
+
+from unsloth import FastLanguageModel
+import torch
+
+max_seq_length = 2048 # Choose any! We auto support RoPE Scaling internally!
+dtype = None # None for auto detection. Float16 for Tesla T4, V100, Bfloat16 for Ampere+
+load_in_4bit = True # Use 4bit quantization to reduce memory usage. Can be False.
+
+alpaca_prompt = """Below is an instruction that describes a task, along with an input that provides additional context. Write a response that appropriately completes the request.
+
+### Instruction:
+{}
+
+### Input:
+{}
+
+### Response:
+{}"""
+
+class Question(BaseModel):
+    query: str
+
+@router.post("/generate_answer")
+def generate_answer(value: Question):
+    try:
+        llama_model, llama_tokenizer = FastLanguageModel.from_pretrained(
+            model_name = "Antonio27/llama3-8b-4-bit-for-sugar",
+            max_seq_length = max_seq_length,
+            dtype = dtype,
+            load_in_4bit = load_in_4bit,
+        )
+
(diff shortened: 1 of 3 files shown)


chat/router.py
```diff
@@ -0,0 +1,110 @@
+from fastapi import APIRouter
```
chimosky, line 1: This is imported but unused.

chat/router.py
```diff
@@ -0,0 +1,110 @@
...
+                    f'''
+                    Your task is to answer children's questions using simple language.
+                    Explain any difficult words in a way a 3-year-old can understand.
+                    Keep responses under 60 words.
+                    \n\nQuestion: {value.query}
+                    ''',  # instruction
+                    "",  # input
+                    "",  # output - leave this blank for generation!
+                )
+            ], return_tensors="pt").to("cuda")
```
chimosky, line 58: These should be assigned once and used whenever as it seems their params won't change after the first assignment, there's no reason to reassign them whenever the function is called.

chat/router.py
```diff
@@ -0,0 +1,110 @@
...
+### Input:
+{}
+
+### Response:
+{}"""
+
+class Question(BaseModel):
+    query: str
+
+@router.post("/generate_answer")
```
chimosky, line 25: `router` doesn't exist yet, how come you're using it?

chimosky:
Reviewed ce06570, your commit message should be improved to explain what exactly we're doing with FastAPI and why.

The header is just a summary, your change needs more than a summary.

chat/router.py
XXXJumpingFrogXXX: I modified the code in PR # 9. This is used in the new code and essential for fastapi part.

chat/router.py
XXXJumpingFrogXXX: Here's the above question. I have modified that.

XXXJumpingFrogXXX:
> Reviewed [ce06570](https://github.com/sugarlabs/sugar-ai/commit/ce0657020b32065176e78fbe3f02cb9c34f152b7), your commit message should be improved to explain what exactly we're doing with FastAPI and why.
> 
> The header is just a summary, your change needs more than a summary.
> 
> Also, please avoid opening PRs with your master branch, checkout our [contributing](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#contributing) docs.

I agree. I make a detailed introduction in the new PR #9

XXXJumpingFrogXXX:
I made the latest modifications in PR #9 and provided a detailed introduction, so I will close this PR.

PR #11, opened 2024-11-15
Pranshul

Pranshul-Thakur:
> setup an error handling procedure in setup_vectorstore
> also setup an error Handling in get_relevant_document
> and added warning for missing files

### DIFF
--- a/rag_agent.py
+++ b/rag_agent.py
@@ -33,7 +33,6 @@ Question: {question}
 Answer: Let's think step by step.
 """
 
-
 class RAG_Agent:
     def __init__(self, model="llama3.1"):
         self.model = OllamaLLM(model=model)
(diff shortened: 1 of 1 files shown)


chimosky:
There's no need to setup the error handling you're trying to setup, what did you notice when trying to test this and can you share steps to reproduce?

Pranshul-Thakur:
i added the error handling part for better clarity while trying to run it and the output i was getting it while trying it out was as follows 
image
prior to this while running the original one, i kept getting undefined errors

chimosky:
An undefined errror could means the package import didn't work, maybe because you didn't have the dependency installed.

In the case of the image you've shared above - you could share text next time as that's clearer -, `qa_chain` is created on [L109](https://github.com/sugarlabs/sugar-ai/blob/main/rag_agent.py#L109) before it's used anywhere.

@kshitijdshah99 can you reproduce this?

Pranshul-Thakur:
ok, i ll try reinstalling it all and doing it again, and get back one i got it working

kshitijdshah99:
> An undefined errror could means the package import didn't work, maybe because you didn't have the dependency installed.
> 
> In the case of the image you've shared above - you could share text next time as that's clearer -, `qa_chain` is created on [L109](https://github.com/sugarlabs/sugar-ai/blob/main/rag_agent.py#L109) before it's used anywhere.
> 
> @kshitijdshah99 can you reproduce this?

Yes even I feel the same maybe there's some issue with the dependencies or their incompatibility. Bcz when I had worked with with it the model was working great. It's the langchain dependencies that are causing this trouble.

kshitijdshah99:
> `qa_chain` is created on [L109](https://github.com/sugarlabs/sugar-ai/blob/main/rag_agent.py#L109) before it's used anywhere.
> 
> @kshitijdshah99 can you reproduce this?

You mean that I initialize the qa_chain only after the query comes in and the document is retrieved?

chimosky:
> the undefined error was because of connection issue, sorry for that

Thanks for the update, closing this now as it's not something we want.

PR #14, opened 2025-01-19
Update dependencies

MostlyKIGuess:
- security fixes

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 absl-py==1.4.0
 accelerate==0.34.2
 aiohappyeyeballs==2.4.0
-aiohttp==3.10.5
+aiohttp==3.10.11
 aiosignal==1.3.1
 alabaster==0.7.16
 albucore==0.0.15
@@ -55,20 +55,20 @@ debugpy==1.6.6
 decorator==4.4.2
 defusedxml==0.7.1
 
-Jinja2==3.1.4
+Jinja2==3.1.5
 joblib==1.4.2
 jsonpickle==3.3.0
 jsonschema==4.23.0
 jsonschema-specifications==2023.12.1
 jupyter-client==6.1.12
 jupyter-console==6.1.0
-jupyter-server==1.24.0
+jupyter-server==2.7.2
 jupyter_core==5.7.2
 jupyterlab_pygments==0.3.0
 jupyterlab_widgets==3.0.13
 kaggle==1.6.17
 kagglehub==0.2.9
-keras==3.4.1
+keras==3.8.0
 keyring==23.5.0
 kiwisolver==1.4.7
 langcodes==3.4.0
@@ -150,7 +150,7 @@ tzlocal==5.2
 uc-micro-py==1.0.3
 unsloth @ git+https://github.com/unslothai/unsloth.git@62c989ef0ae0e9fbac714a4cb21eda76c1fe84b6
 uritemplate==4.1.1
-urllib3==2.0.7
+urllib3==2.2.2
 vega-datasets==0.9.0
 wadllib==1.3.6
 wasabi==1.1.3
@@ -159,7 +159,7 @@ weasel==0.4.1
 webcolors==24.8.0
 webencodings==0.5.1
 websocket-client==1.8.0
-Werkzeug==3.0.4
+Werkzeug==3.0.6
 widgetsnbextension==3.6.9
 wordcloud==1.9.3
 wrapt==1.16.0


chimosky:
Did you test your changes to be sure nothing breaks with these versions?

MostlyKIGuess:
> Did you test your changes to be sure nothing breaks with these versions?

while checking I realized , most of them are not needed, I only kept the necessary ones , will push the commit for that. I saw the mail and updated dependencies, want to fix this before this tuesday.

MostlyKIGuess:
Welp the master branch wasn't working, so I made a lot of changes, almost revamping. But shouldn't we just merge the documents of this along with https://sugar-docs-ai.streamlit.app/. But I am willling to update this as well. I just saw the mail regarding dep issues and thought would fix it.

MostlyKIGuess:
issue with current master is that ollama and httpx can't be run simultaneity they break each other's dependencies.

chimosky:
> issue with current master is that ollama and httpx can't be run simultaneity they break each other's dependencies.

Are you referring to `aiohttp` and `ollama` or `langchain-ollama`? 

Also @kshitijdshah99 why do we have two ollama dependencies?

MostlyKIGuess:
> > issue with current master is that ollama and httpx can't be run simultaneity they break each other's dependencies.
> 
> Are you referring to `aiohttp` and `ollama` or `langchain-ollama`?
> 
> Also @kshitijdshah99 why do we have two ollama dependencies?

ollama and httpx, check https://github.com/ollama/ollama-python/issues/356

kshitijdshah99:
> > issue with current master is that ollama and httpx can't be run simultaneity they break each other's dependencies.
> 
> Are you referring to `aiohttp` and `ollama` or `langchain-ollama`?
> 
> Also @kshitijdshah99 why do we have two ollama dependencies?

Both these dependencies serve different purpose. ```langchain-ollama``` is a package which provides interaction between Langchain and Ollama to build Langchain workflows. On other hand ```ollama``` is for extracting and managing different models like llama3.1, Mistral or Codellama we downloaded locally on our system.

MostlyKIGuess:
> > > issue with current master is that ollama and httpx can't be run simultaneity they break each other's dependencies.
> > 
> > Are you referring to `aiohttp` and `ollama` or `langchain-ollama`?
> > 
> > Also @kshitijdshah99 why do we have two ollama dependencies?
> 
> Both these dependencies serve different purpose. ```langchain-ollama``` is a package which provides interaction between Langchain and Ollama to build Langchain workflows. On other hand ```ollama``` is for extracting and managing different models like llama3.1, Mistral or Codellama we downloaded locally on our system. 

We should just pull from HF instead this way we can remove ollama. What say?

chimosky:
> I'll suggest if @chimosky agrees we can try solving this issue first instead of directly skipping to HuggingFace because then we would have to adjust many code files.

If we can do without a dependency, then we should.

requirements.txt
```diff
@@ -1,187 +1,13 @@
...
-yarl==1.11.1
-yellowbrick==1.5
-yfinance==0.2.43
-zict==3.0.0
-zipp==3.20.1
-
-# Rag agent requirments
 langchain
+transformers==4.45.2
 langchain-ollama
```
harshkasat, line 3: can you remove langchain-ollama because we're not using ollama in his pr

rag_agent.py
```diff
@@ -112,25 +102,50 @@ def format_docs(docs):
...
+            if not question:
+                print("Please enter a valid question.")
+                continue
+            response = agent.run(question)
+            print("Response:", response)
+    except Exception as e:
+        print(f"An error occurred: {e}")
+
+if __name__ == "__main__":
+    main()
```
chimosky, line 151: You should remove the No EOF at end of file.

rag_agent.py
```diff
@@ -112,25 +102,50 @@ def format_docs(docs):
...
 
+        while True:
+            question = input("Enter your question: ").strip()
+            if not question:
+                print("Please enter a valid question.")
+                continue
+            response = agent.run(question)
+            print("Response:", response)
+    except Exception as e:
+        print(f"An error occurred: {e}")
```
chimosky, line 148: Considering this is an API, we definitely wouldn't use command line args for it.

chimosky:
> Are there any criteria for open-source models? I'm curious, especially since new models like DeepPeek and Qwen perform well in coding and mathematics with their instruction models.

The only thing we cared about was how the model responded based on the prompt, which was adjusted of course.

kshitijdshah99:
@MostlyKIGuess I saw you are using very smaller models from HF which is a little less preferable....and of course needless to say that 8B models are taking lot of time for inference so it becomes necessary to use smaller ones. Have you find out some middle way to solve this problem? I tried running Llama-8B model from HF but it's consuming lot's of RAM and space too while running.

kshitijdshah99:
@chimosky it seems these guys have fixed this interdependency issue. See the latest [messages](https://github.com/ollama/ollama-python/issues/356). Probably we can work with the latest httpx version if we want to.

MostlyKIGuess:
> @MostlyKIGuess I saw you are using very smaller models from HF which is a little less preferable....and of course needless to say that 8B models are taking lot of time for inference so it becomes necessary to use smaller ones. Have you find out some middle way to solve this problem? I tried running Llama-8B model from HF but it's consuming lot's of RAM and space too while running.

I did that just to test the pipeline, by no means I promote using that and your point of using ollama isn't  in good faith, as ollama using quantization on the model to run it, the models in themselves remain the same, we can do the same with huggingface, I am by no means in support of huggingface, it's just convenient because we get access to more models, with ollama there's a dep hole. I will be working on the reviews today and try to remove more bloat.

MostlyKIGuess:
I also noticed;
```
  from langchain.embeddings import HuggingFaceEmbeddings
Hardware accelerator e.g. GPU is available in the environment, but no `device` argument is passed to the `Pipeline` object. Model will be on CPU.
```

MostlyKIGuess:
I will be fixing that shortly today

kshitijdshah99:
> > @MostlyKIGuess I saw you are using very smaller models from HF which is a little less preferable....and of course needless to say that 8B models are taking lot of time for inference so it becomes necessary to use smaller ones. Have you find out some middle way to solve this problem? I tried running Llama-8B model from HF but it's consuming lot's of RAM and space too while running.
> 
> I did that just to test the pipeline, by no means I promote using that and your point of using ollama isn't in good faith, as ollama using quantization on the model to run it, the models in themselves remain the same, we can do the same with huggingface, I am by no means in support of huggingface, it's just convenient because we get access to more models, with ollama there's a dep hole. I will be working on the reviews today and try to remove more bloat.

Bro ik and am convinced with your purpose of opening this PR. The prime reason for supporting Ollama is bcz I have integrated it with fastAPI to the UI and also it's faster for the inference that's why I am not preferring HF. The primary reason of dependency conflict is not there anymore hence I don't feel the need of migrating the code. That's my point. Ollama like HF has many open source models (including deepseek) so that's not going to be an issue in future.

kshitijdshah99:
> I also noticed;
> 
> ```
>   from langchain.embeddings import HuggingFaceEmbeddings
> Hardware accelerator e.g. GPU is available in the environment, but no `device` argument is passed to the `Pipeline` object. Model will be on CPU.
> ```

I preferred using CPU bcz I hadn't one in my system yes we can improve it.

harshkasat:
I will suggest using collab if you want to leverage GPU or make ssh or tunnel to use GPU in vs code

MostlyKIGuess:
> I will suggest using collab if you want to leverage GPU or make ssh or tunnel to use GPU in vs code

I have resources actually personally to test it out.

MostlyKIGuess:
> are there any TODO fixes ? I also want to work on this.

maybe review the changes? will push in like 5-10 mins? would be a great help

MostlyKIGuess:
Getting a 9.96 gb Image now with total GPU support!

MostlyKIGuess:
> with model?

no no, this is just the image generated container, model would add space depending upon which we use, I have commented 2 and kept the lowest parameter one because I wanted to just test the pipeline.

MostlyKIGuess:
https://hub.docker.com/repository/docker/mostlyk/sugar-ai/general

harshkasat:
I also find some low-level design issues:
1) Single Responsibility Principle: since RagAgent does a lot of task embed, storing, generate, I think we can create a base class and follow abstraction, 
2) Dependency Injection Violation: `agent.retriever = agent.setup_vectorstore(args.docs)` modifidying retriever object after creating.
there might be other issues but I find this ones.

MostlyKIGuess:
> I also find some low-level design issues:
> 
>     1. Single Responsibility Principle: since RagAgent does a lot of task embed, storing, generate, I think we can create a base class and follow abstraction,
> 
>     2. Dependency Injection Violation: `agent.retriever = agent.setup_vectorstore(args.docs)` modifidying retriever object after creating.
>        there might be other issues but I find this ones.

For this, I haven't actually worked mainly on this project , I did make another similar project it is called sugar docs ai in my repositories, well at least as of now, the immediate action was to lower the size and fix the dep loophole, perhaps we can create a ticket and start working on these too?

chimosky:
I haven't reviewed your changes yet, please write better commit messages, the ones I'm seeing above aren't helpful.

This PR was originally for updating dependencies and it's become more than that, how about you open a new PR with the changes that have nothing to do with updating dependencies and write better commit messages.

MostlyKIGuess:
> I haven't reviewed your changes yet, please write better commit messages, the ones I'm seeing above aren't helpful.
> 
> This PR was originally for updating dependencies and it's become more than that, how about you open a new PR with the changes that have nothing to do with updating dependencies and write better commit messages.

alright, will close this then!

Issue #20, opened 2025-03-05
Improve RAG Agent response.


chimosky:
@MostlyKIGuess  This is to track the improvement to the response which potentially includes modifying the prompts.

imshrishk:
I can work on this as I have developed and fine-tuned a RAG model before. Just want to make sure what is the metric we are using to figure out the improvement. Unless we have any new data to train on, I think the only option is prompt engineering.

chimosky:
Please look at #17  to see the discussion that's happened so far.

Khan-Ramsha:
Hello @chimosky , I would like to take up this issue. Have gone through the convo y'all had in this pr #17 , I’ve been working on improving the model’s responses, and I think I’ve cracked it. Will be raising a PR in a bit!

aazeeem11:
Hi @chimosky,

Based on the discussion in #17 and this issue, I made a small improvement focused on response clarity for children (ages 5–12).

The change is limited to prompt wording only and aims to reduce verbosity and self-talk, while keeping explanations simple and friendly. Optional real-life analogies are used only when helpful.

I tested the update locally using the /ask endpoint.

A PR is opened with this prompt-only change. 

Thanks!

PR #17, opened 2025-02-07
Docker Support and Improve RAG_Agent Implementation

MostlyKIGuess:
## Summary  
This PR introduces a **Dockerfile** to containerize Sugar-AI, ensuring a consistent and reproducible environment. Additionally, it modifies the **RAG_Agent** by refactoring its code, addressing dependency issues, and adding unit tests.

## Changes  
- **Added Dockerfile** for containerization.
- **Updated README** with instructions for running Sugar-AI using Docker.  
- **Refactored RAG_Agent** to:
  - Utilize **Hugging Face (HF) models** for better performance.  
  - Fix **Ollama dependency issues**.  
  - Update relevant **dependencies** for compatibility.  
- **Added unit tests** to validate RAG_Agent functionality.

PR #23, opened 2025-04-05
feat: improve rag responses with better prompt and gen params.

Khan-Ramsha:
Fixes for #20 

Went through all the discussion in PR #17, 
here's what i got--> Currently, there are two endpoints — /ask and /ask-llm — and we’re focusing on the former.
The /ask endpoint tries to fetch document content related to the user query and uses that to answer the question. But there’s a twist: it runs through two pipelines — the first one generates the answer, and the second one simplifies that answer for kids. Then the final simplified response is returned to the user.

But here’s the problem:
The model gives different responses every time for the same exact query — super inconsistent. Also, it doesn’t seem to actually use the document content for answering the question. On top of that, it sometimes cuts off the response halfway.

And when I tested the current main branch code, the model was generating super vague answers, kind of like it was “talking to itself.” It kept repeating phrases and even echoed the user’s prompt.

Here's how i fixed all that--> When i peeked into the code, turns out the model wasn’t even getting any document content tied to the user’s query. So how could it use something it never got?
To fix that, I included the document content directly in the prompt — now the model actually has context to work with. Next, to stop the “self-talking” behavior, I adjusted the prompt to explicitly tell the model not to repeat or talk to itself — that helped cut out the weirdness. Since the model’s behavior was non-deterministic (random), I set do_sample = False to make the outputs consistent every time. Also added: temperature, top_k, top_p = None , these settings basically kills the randomness, and randomness is what we don’t want here.

Then, I used return_full_text = False so we only get the generated output (not the full prompt + output). This stopped the model from repeating parts of the prompt. And repetition_penalty helps prevent repetitive phrases in the response.

And yep — added max_new_tokens to control response length. Without it, the model either cuts off early or overshoots and starts rambling. This keeps answers clean and to the point.

Lastly, I added a trim_incomplete_sentences() function to clean up outputs that were being cut off mid-sentence — now responses end properly.

All of this genuinely improved the model’s responses.
They now stick to the document content and the final output is clear, consistent, and simplified for kids.

### DIFF
--- a/rag_agent.py
+++ b/rag_agent.py
@@ -2,6 +2,7 @@
 # Uses a model from HuggingFace with optional 4-bit quantization
 
 import os
+import re
 import argparse
 import torch
 from transformers import pipeline
@@ -12,75 +13,68 @@ from langchain_community.document_loaders import PyMuPDFLoader, TextLoader
 from langchain_core.runnables import RunnablePassthrough
 from langchain.prompts import ChatPromptTemplate
 
-
 PROMPT_TEMPLATE = """
-You are a highly intelligent Python coding assistant built for kids using the Sugar Learning Platform.
-1. Focus on coding-related problems, errors, and explanations.
-2. Use the knowledge from the provided Pygame, GTK, and Sugar Toolkit documentation.
-3. Provide complete, clear and concise answers.
-4. Your answer must be easy to understand for kids.
-5. Always include Sugar-specific guidance when relevant to the question.
+You are a smart and helpful assistant designed to answer coding questions using the Sugar Learning Platform.
+Instructions:
+1. You must ONLY use the information from the provided context to answer the question.
+2. If the context doesn't fully answer the question, use your general knowledge to provide the most helpful and accurate answer you can."
+3. Do NOT mention the context, documents, or how the answer was generated. Just provide the answer naturally and clearly.
+4. When possible, prioritize and include any relevant details from the context.
+5. Always answer in a concise, accurate, and helpful manner.
 
+Context: {context}
 Question: {question}
+
 Answer:
 """
 
 CHILD_FRIENDLY_PROMPT = """
-Your task is to answer children's questions using simple language.
-You will be given an answer, you will have to paraphrase it.
-Explain any difficult words in a way a 5-12-years-old can understand.
+You are a friendly teacher talking to a child aged 3 to 10 years old.
 
-Original answer: {original_answer}
+Rewrite the answer below using simple words and short sentences so a young child can understand it.
 
-Child-friendly answer:
+Include examples if needed. Stay close to the original meaning. Do not add extra commentary or explanation about what you are doing. .
+
+Here is the answer to simplify:
+{original_answer}
+
 """
 
-
+def trim_incomplete_sentence(text):
+    matches = list(re.finditer(r'\.\s', text))
+    if matches:
+        last_complete = matches[-1].end()
+        return text[:last_complete].strip()
+    else:
+        return text.strip()
 
 def format_docs(docs):
-    """Return all document content separated by two newlines."""
     return "\n\n".join(doc.page_content for doc in docs)
 
-
 def combine_messages(x):
-    """
-    If 'x' has a method to_messages, combine message content with newline.
-    Otherwise, return string representation.
-    """
     if hasattr(x, "to_messages"):
         return "\n".join(msg.content for msg in x.to_messages())
     return str(x)
 
-
 def extract_answer_from_output(outputs):
-    """
-    Extract the answer text from the model's output after the keyword 'Answer:'.
-    """
     generated_text = outputs[0]['generated_text']
-
     if "Child-friendly answer:" in generated_text:
-        return generated_text.split("Child-friendly answer:")[-1].strip()
-    
+        return generated_text.split("Here is the answer to simplify:")[-1].strip()
     return generated_text.split("Answer:")[-1].strip()
 
-
 class RAG_Agent:
-    def __init__(self, model="Qwen/Qwen2-1.5B-Instruct",
-                 quantize=True):
-        # Disable quantization if CUDA is not available
+    def __init__(self, model="Qwen/Qwen2-1.5B-Instruct", quantize=True):
         self.use_quant = quantize and torch.cuda.is_available()
         self.model_name = model
-        
+
         if self.use_quant:
             from transformers import AutoModelForCausalLM, AutoTokenizer, BitsAndBytesConfig
-
             bnb_config = BitsAndBytesConfig(
                 load_in_4bit=True,
                 bnb_4bit_compute_dtype=torch.float16,
                 bnb_4bit_use_double_quant=True,
                 bnb_4bit_quant_type="nf4"
             )
-
             tokenizer = AutoTokenizer.from_pretrained(model)
             model_obj = AutoModelForCausalLM.from_pretrained(
                 model,
@@ -88,66 +82,15 @@ class RAG_Agent:
                 torch_dtype=torch.float16,
                 device_map="auto"
             )
-            self.model = pipeline(
-                "text-generation",
-                model=model_obj,
-                tokenizer=tokenizer,
-                max_length=1024,
-                truncation=True,
-            )
-            
-            tokenizer2 = AutoTokenizer.from_pretrained(model)
-            self.simplify_model = pipeline(
-                "text-generation",
-                model=model_obj,  
-                tokenizer=tokenizer2,
-                max_length=1024,   
-                truncation=True,
-            )
+            self.model = pipeline("text-generation", model=model_obj, tokenizer=tokenizer, max_new_tokens=512, return_full_text=False, do_sample=False, temperature=None, top_p=None, top_k=None, repetition_penalty=1.2, truncation=True)
+            self.simplify_model = self.model
         else:
-            self.model = pipeline(
-                "text-generation",
-                model=model,
-                max_length=1024,
-                truncation=True,
-                torch_dtype=torch.float16,
-                device=0 if torch.cuda.is_available() else -1,
-            )
-
-            self.simplify_model = pipeline(
-                "text-generation",
-                model=model,
-                max_length=1024,
-                truncation=True,
-                torch_dtype=torch.float16,
-                device=0 if torch.cuda.is_available() else -1,
-            )
+            self.model = pipeline("text-generation", model=model, max_new_tokens=512, return_full_text=False, do_sample=False, temperature=None, top_p=None, top_k=None, repetition_penalty=1.2, truncation=True, torch_dtype=torch.float16, device=0 if torch.cuda.is_available() else -1)
+            self.simplify_model = self.model
 
         self.retriever = None
         self.prompt = ChatPromptTemplate.from_template(PROMPT_TEMPLATE)
         self.child_prompt = ChatPromptTemplate.from_template(CHILD_FRIENDLY_PROMPT)
-
-    def set_model(self, model):
-        # Update both models
-        self.model_name = model
-        self.model = pipeline(
-            "text-generation",
-            model=model,
-            max_length=1024,
-            truncation=True,
-            torch_dtype=torch.float16
-        )
-        
-        self.simplify_model = pipeline(
-            "text-generation",
-            model=model,
-            max_length=1024,
-            truncation=True,
-            torch_dtype=torch.float16
-        )
-
-    def get_model(self):
-        return self.model
 
     def setup_vectorstore(self, file_paths):
         all_documents = []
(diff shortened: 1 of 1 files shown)


MostlyKIGuess:
There seems to be changes which aren't needed, all we had to do for that issue to re-write the prompt , could you also share the responses as discussed at the end in #17  on questions.

Khan-Ramsha:
Results Before (i.e current main branch):
image

image

image

Results After( after my changes):

Prompt passing to model for simplifying response-->>
image

Final Response
image

Prompt passing to model for simplifying response-->>

image

Final Response

image

Khan-Ramsha:
> There seems to be changes which aren't needed, all we had to do for that issue to re-write the prompt , could you also share the responses as discussed at the end in #17 on questions.

Yep, the main issue was with the prompt. But I also found that tweaking some other parameters like temperature and repetition penalty helped improve the model's responses overall. It reduced repetition and made the answers simpler and more consistent, especially for the child-friendly prompt.

Khan-Ramsha:
@chimosky have a look!

MostlyKIGuess:
> But here’s the problem: The model gives different responses every time for the same exact query — super inconsistent. Also, it doesn’t seem to actually use the document content for answering the question. On top of that, it sometimes cuts off the response halfway.

No, this is not a problem, the answer is usually dependent on the seed you are on, you can potentially set seed = -1 (prev) or 0, for consistent responses. That is how LLM work. And it does retrieve information on /ask, we can set the embeddings to show later but we don't need it right now. Also even if it did not I do not see what part of the change would make it work.

> And when I tested the current main branch code, the model was generating super vague answers, kind of like it was “talking to itself.” It kept repeating phrases and even echoed the user’s prompt.
> 
> Here's how i fixed all that--> When i peeked into the code, turns out the model wasn’t even getting any document content tied to the user’s query. So how could it use something it never got? To fix that, I included the document content directly in the prompt — now the model actually has context to work with. Next, to stop the “self-talking” behavior, I adjusted the prompt to explicitly tell the model not to repeat or talk to itself — that helped cut out the weirdness. Since the model’s behavior was non-deterministic (random), I set do_sample = False to make the outputs consistent every time. Also added: temperature, top_k, top_p = None , these settings basically kills the randomness, and randomness is what we don’t want here.
> 
> Then, I used return_full_text = False so we only get the generated output (not the full prompt + output). This stopped the model from repeating parts of the prompt. And repetition_penalty helps prevent repetitive phrases in the response.

This can be useful but we can decide on how we want to keep it, also that didn't stop your model from repeating, it was the max tokens that you reduced.

> And yep — added max_new_tokens to control response length. Without it, the model either cuts off early or overshoots and starts rambling. This keeps answers clean and to the point.

This was discussed in #17  on what to keep, we ended up keeping 1024 because we need enough length to go through some topics.

Khan-Ramsha:
> No, this is not a problem, the answer is usually dependent on the seed you are on, you can potentially set seed = -1 (prev) or 0, for consistent responses. That is how LLM work.

Yes, the model can give different outputs for the same query depending on the seed, which is pretty normal for LLMs.  but since we have do_sample=False, the model should already be generating deterministic responses.

Khan-Ramsha:
> And it does retrieve information on /ask, we can set the embeddings to show later but we don't need it right now.
I agree, Got it! The context is being directly passed in the prompt template so the model has the information upfront. As for the embeddings, yes they can be shown later if needed

Khan-Ramsha:
> This was discussed in #17 on what to keep, we ended up keeping 1024 because we need enough length to go through some topics.

Ohh, will increase it to 1024 then

Khan-Ramsha:
> Suggestions: 1.The code can be refactored to just changing the prompt. 2. As discussed in #17 the reason we wrote everything on new line was because we want to code to be readable so all the changes on functions aren't required. 3. Max_Token is something we need to discuss and can keep repetition penalty and DoLa layers.

Changes Done, 

Results:

image

Response:-

image

image

Response:-
image

All set , take a look!

rag_agent.py
```diff
@@ -92,23 +105,41 @@ def __init__(self, model="Qwen/Qwen2-1.5B-Instruct",
...
             )
             
             tokenizer2 = AutoTokenizer.from_pretrained(model)
             self.simplify_model = pipeline(
                 "text-generation",
                 model=model_obj,  
                 tokenizer=tokenizer2,
-                max_length=1024,   
-                truncation=True,
+                max_new_tokens=512,
```
MostlyKIGuess, line 123: 1024 right?

rag_agent.py
```diff
@@ -92,23 +105,41 @@ def __init__(self, model="Qwen/Qwen2-1.5B-Instruct",
...
+                top_k=None, 
+                repetition_penalty=1.2, 
+                truncation=True
             )
         else:
             self.model = pipeline(
                 "text-generation",
                 model=model,
-                max_length=1024,
+                max_new_tokens=512,
```
MostlyKIGuess, line 136: here as well

rag_agent.py
```diff
@@ -133,15 +170,27 @@ def set_model(self, model):
...
+            repetition_penalty=1.2, 
             truncation=True,
             torch_dtype=torch.float16
         )
         
         self.simplify_model = pipeline(
             "text-generation",
             model=model,
-            max_length=1024,
+            max_new_tokens=512,
```
MostlyKIGuess, line 187: .

rag_agent.py
```diff
@@ -133,15 +170,27 @@ def set_model(self, model):
         self.model = pipeline(
             "text-generation",
             model=model,
-            max_length=1024,
+            max_new_tokens=512,
```
MostlyKIGuess, line 173: .

rag_agent.py
```diff
@@ -167,52 +216,80 @@ def setup_vectorstore(self, file_paths):
...
+                if "query" in query:
+                    query = query["query"]
+                else:
+                    query = next(iter(query.values()))
+            
+            query = str(query)
+            
+            results = self.retriever.invoke(query)
+            
+            print(f"[DEBUG] Retrieved results: {results}")
```
MostlyKIGuess, line 230: Thank you, this is great for debugging purposes, maybe add it as an extra commit and mention this, better to keep the repo documented well

rag_agent.py
```diff
@@ -257,7 +335,9 @@ def main():
             print("Response:", response)
     except Exception as e:
         print(f"An error occurred: {e}")
+        import traceback
+        traceback.print_exc()
 
 
 if __name__ == "__main__":
-    main()
+    main()
```
MostlyKIGuess, line 343: eof missing

rag_agent.py
```diff
@@ -223,7 +300,8 @@ def main():
         choices=[
             'bigscience/bloom-1b1',
             'facebook/opt-350m',
-            'EleutherAI/gpt-neo-1.3B'
+            'EleutherAI/gpt-neo-1.3B',
+            'Qwen/Qwen2-1.5B-Instruct'
```
MostlyKIGuess, line 304: not needed

rag_agent.py
```diff
@@ -167,52 +216,80 @@ def setup_vectorstore(self, file_paths):
...
+                self.prompt  
+                | print_prompt_before_model
+                | self.model
+                | extract_answer_from_output
+            )
+            
             first_response = first_chain.invoke({
-                "query": question,
-                "context": doc_result.page_content
+                "question": question,
```
MostlyKIGuess, line 265: again , I don't see how this change is required here 245-292

MostlyKIGuess (changes requested):
Reviewed yet to test

rag_agent.py
Khan-Ramsha: Aah alright, changing it then! I thought we were supposed to figure out first whether we’re sticking with max_new_tokens or max_length, so I was waiting to change it accordingly.

rag_agent.py
Khan-Ramsha: > again , I don't see how this change is required here 245-292

Chatprompttemplate expects key to be 'question', not query. As in the prompt i am passing '{question}'

Khan-Ramsha: So if am changing this key from 'question' to 'query ' again i need to do changes in the Prompt too

rag_agent.py
Khan-Ramsha: Alright, did the changes

rag_agent.py
Khan-Ramsha: Okay so i need to remove this line for now, then adding it again with a proper commit message!

rag_agent.py
Khan-Ramsha: Done

rag_agent.py
Khan-Ramsha: done

rag_agent.py
Khan-Ramsha: Done

chimosky:
Haven't tested or looked at it deeply, but I agree with @MostlyKIGuess.

There's way too many unnecessary changes.

Khan-Ramsha:
> Haven't tested or looked at it deeply, but I agree with @MostlyKIGuess.
> 
> There's way too many unnecessary changes.
> 
> Also see [making commits](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#making-commits) to write better commit messages.

Fair point! I agree with @MostlyKIGuess too. I went a bit overboard and ended up changing whole codebase rather than just focusing on prompts which would have been enough solve the issue. That said, my lastest commit includes only prompt engineering and model params tweaking. Pls take a moment to review it and lemme know if you have any suggestions or changes.

chimosky:
You're yet to push said latest commits.

Khan-Ramsha:
> You're yet to push said latest commits.

I had already pushed the latest commits before asking @MostlyKIGuess for review. lemme know if anything's still missing, I will make the changes.

chimosky:
Looking at the changes you've made, nothing has changed.

Khan-Ramsha:
> Looking at the last changes you pushed, nothing has changed.
> 
> I'd made my comments after looking at those changes, I'm wondering why you're saying you pushed changes when I don't see any.

Correct, I did not push anything after this commit: [7fd33c6](https://github.com/sugarlabs/sugar-ai/pull/23/commits/7fd33c625d562ed9800d40402cb5ca4b4bee9755) 

The only code I changed was refining the prompt by passing the full document content to it, and adding a few model parameters.
I’ve tested the updated prompt along with the model parameters, but I think the prompt already guides the model clearly enough, so the extra model parameters might be unnecessary.

Should I go ahead and remove them?
If you were pointing to any other unnecessary changes,  could you please point them out? I’ll clean those up too.

Thankyou for your guidance. I really appreciate it and want to make sure I’m aligned with what’s expected.

Khan-Ramsha:
> so the extra model parameters might be unnecessary.


image

These are the ones I meant, as I believe the prompt already guides the model well enough in that direction. Still, I'll test it once by removing these parameters to see if there's any noticeable difference.

Khan-Ramsha:
> > Should I go ahead and remove them?
> 
> I figured that was the unspoken agreement considering we'd said it was unnecessary.

Got it, sir! removing all unnecessary changes.

Khan-Ramsha:
Just noticed temperature and a few others didn’t show up properly in the commit message — removed do_sample, temperature, top_p, top_k, repetition_penalty, and return_full_text for cleaner defaults. Missed listing them all due to formatting, sorry about that!

Khan-Ramsha:
Done! I tried to improve the model responses by making only the necessary changes. Have a look, sir, and let me know if anything still needs to be updated.

rag_agent.py
```diff
@@ -184,82 +184,50 @@ def setup_vectorstore(self, file_paths):
...
-            final_response = second_chain.invoke({
-                "original_answer": first_response
-            })
-            
-            return trim_incomplete_sentence(final_response)
-            
-        except Exception as e:
-            print(f"Error in run method: {e}")
+        if not context_text.strip():
+            return "I couldn't find an answer in the documents."
```
chimosky, line 213: Not all answers are supposed to be in the documents, the documents are there for Sugar specific context not as the sole source of answers.

chimosky:
Your commit message in [396edd4](https://github.com/sugarlabs/sugar-ai/pull/23/commits/396edd4cad48cf5da4133bd3677de592825af183), I'm guessing you meant parsing and not passing.

Would also be nice if you didn't use all the width provided by your editor, makes the commit message easier to read.

chimosky (commented):
Debug prints shouldn't be part of your changes as they're not useful to anyone else but you while you were debugging, please don't include them.

chimosky:
@MostlyKIGuess could you review and test this, I can't test at the moment.

rag_agent.py
Khan-Ramsha: Totally agree the model should still answer even if there's no doc context. That was a temp check I forgot to remove. Will fix it

Khan-Ramsha:
> Your commit message in [396edd4](https://github.com/sugarlabs/sugar-ai/pull/23/commits/396edd4cad48cf5da4133bd3677de592825af183), I'm guessing you meant parsing and not passing.

I actually meant "passing". Just feeding the formatted context straight into the prompt

Khan-Ramsha:
> Would also be nice if you didn't use all the width provided by your editor, makes the commit message easier to read.

Alright, will keep that in mind for future commits

Khan-Ramsha:
> Debug prints shouldn't be part of your changes as they're not useful to anyone else but you while you were debugging, please don't include them.

Sure. I will make sure to remove those before I push my changes to the repository.

PR #32, opened 2025-09-16
Webhook based CI/CD

mebinthattil:
I have tested this and it works.

---

Once GH fires a webhook and it is validated, it does a hard reset to avoid conflicts if any. 
It then proceeds to restart the sugarai service, this service takes around 10-15 seconds to start up again. Users may experience this 10-15seconds downtime.


chimosky:
Reviewed, not tested.

Please add the license excerpt.

PR #29, opened 2025-08-21
Add `/ask-llm-prompted` endpoint to use custom system prompt and model generation parameters

mebinthattil:
# Description
This PR introduces the /ask-llm-prompted endpoint. It lets you ask the model a question using a custom system prompt and also offers the flexibility to tweak model generation parameters.

Documentation (README) has be updated as well. 

---
### How I tested it:
Tested it using a python script to send a POST request to this endpoint and test it out as the brains for the speak activity. Everything works as expected.

### Files changed:
- `app/ai.py`
- `app/routes/api.py`
- `README.md`

---
Requesting feedback and review from @MostlyKIGuess and @chimosky.

_I would urge that this is put on priority and review it soon, as multiple AI-powered activities need this functionality to set custom system prompts and tweak parameters._

PR #33, opened 2025-09-17
Chat completion endpoint

amannaik247:
This endpoint will serve the conversation part of the story builder feature in the write activity. As well as a general endpoint which can be utilized to generate conversational responses from the gemma model. 

Let's say we are using a gemma model.

- The gemma models are trained using specific tokens to generate model responses using a chat history. More can be understood in this [documentation](https://ai.google.dev/gemma/docs/core/prompt-structure). 
- The main purpose of the endpoint is to take input of the chat history as a list(along with the system prompt in it), like this one:
```
"messages": [
    {"role": "system", "content": "You are a helpful AI assistant.Reply in 2-3 sentences."},
    {"role": "user", "content": "What is machine learning?"},
    {"role": "assistant", "content": "Machine learning is a subset of artificial intelligence..."},
    {"role": "user", "content": "Can you give me a simple example?"}
  ]
```
- Then convert it to LLM compatible format like this one for gemma:
```
<start_of_turn>user
You are a helpful AI assistant.Reply in 2-3 sentences.\n What is machine learning?<end_of_turn>
<start_of_turn>model
Machine learning is a subset of artificial intelligence...<end_of_turn>
<start_of_turn>user
Can you give me a simple example?<end_of_turn>
```
- Using this to generate a reasonable model response.

- This endpoint can be used by any other future activities which want to introduce a conversation related feature as well. Since the endpoint is similar to most popular chat completion endpoints, integration with sugar-ai will become easier.

PS- I tested this locally using a `gemma-3-1b-it` and `Qwen-2-1.5B-Instruct` models.


app/ai.py
```diff
@@ -256,3 +256,102 @@ def run_with_custom_prompt(self, question: str, custom_prompt: str,
...
+            if role == "user" and not first_user_seen and system_content:
+                content = f"{system_content}\n\n{content}"
+                first_user_seen = True
+
+            normalized.append({"role": role, "content": content})
+        return normalized
+
+
+    def _extract_after_prompt(self, full_text: str, prompt: str, eos_token: str = None) -> str:
+        """Return the model completion that comes after the prompt.
```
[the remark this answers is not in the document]
chimosky: Alright, seems a bit strange that the model returns input with output.

app/ai.py
```diff
@@ -257,7 +257,7 @@ def run_with_custom_prompt(self, question: str, custom_prompt: str,
         except Exception as e:
             raise Exception(f"Error generating response with custom prompt: {str(e)}")
 
-    def _normalize_chat_messages(self, messages: list) -> list:
+    def _normalize_chat_messages(self, messages: list[dict]) -> list[dict]:
```
MostlyKIGuess, line 260: Any reason why?

app/routes/api.py
MostlyKIGuess, line 1: Only changes on api.py are just spacing, makes a bit weird to review.. maybe can restore this

mebinthattil:
> > Everyone adding an endpoint for something new isn't a good thing moving forward, this would end up becoming bloated.
> > @MostlyKIGuess We need to figure out a better way to handle this.
> 
> I think @mebinthattil 's custom end point should be perfect for everyone.
> 
> @mebinthattil Could you share maybe a documentation on it on the website? That would be a great PR addition

Yes, the whole point of #29 was so that every activity would not need to raise a PR like this for that activity's usecase. I was in a bit of a rush to get that change out and did not document it well. I shall make a comprehensive documentation about it and also update the website regarding the same. Thanks for pointing this out.

@amannaik247 Try checking out how I implemented this in the speak-ai activity using this generic endpoint.

app/ai.py
amannaik247: Ibiam suggested that it would be better to mention that it is a list of dictionary.

chimosky:
> @MostlyKIGuess @mebinthattil i am using his endpoint for one task. But it only accepts a a custom prompt and a question. There is no way to send a chat history and get a conversation style response. So for providing the context of the whole conversation this endpoint is necessary. Most popular API's have this /chat/completions endpoint.

This makes sense, it'll also be great to modify the endpoint Mebin created to accept this rather than create a whole new endpoint, you can use a default value - set to None - which would have no effect if nothing is passed there.

This would be easier to maintain than a whole new endpoint.

amannaik247:
> @amannaik247 I'm still not convinced why need a separate mode for this. You state that popular APIs have this endpoint and that's why we should also have it. You could achieve the same essential functionality using the `/ask-llm-prompted` endpoint and pass the chat history through that along with an appropriate prompt and model parameters. But if your activity cannot work with that for any reason (I don’t see why it wouldn't ), please state the reason so we can take a call on that.
> 
> This is my two cents, thanks!

If you could go through these links, you would know that any chat model is trained separately using special tokens to do chat completions. The `ask-llm-prompted` endpoint on its own only generates the `Answer` to its query and system prompt. It cannot consider conversation history properly without the use of these special tokens. 

Links:
https://huggingface.co/blog/qwen-3-chat-template-deep-dive
https://ai.google.dev/gemma/docs/gemma_chat#building_the_chatbot 

Additionally, developers can now directly send a list of messages with user and assistant response, then this endpoint converts it into a chat template readable by the LLM. So they don't have to worry about which model is being used and how to decorate their prompt so that the model gives the best response, they just have to use this endpoint! 
Hope this clears your doubt.

amannaik247:
@mebinthattil fixed the older issues and tested the code locally. Let me know if you find any issues after your test. So we can merge this PR.

amannaik247:
@chimosky please check out this PR. I have tested the changes locally.

app/routes/api.py
```diff
@@ -9,16 +9,23 @@
 import os
 import json
 from datetime import datetime
-from typing import Dict, Optional
+from typing import Dict, Optional, List
```
chimosky: Python now provides native support for [typing features](https://typing.python.org/en/latest/guides/modernizing.html) we might have to revisit the use of these later.

PR #43, opened 2026-01-18
prompt: simplify child-friendly responses

aazeeem11:
What changed
- Simplified base and child-friendly prompts to improve clarity for kids (ages 5–12).

Why
- Improves response consistency and readability (Issue #20).

Scope
- Prompt-only change. No logic or parameter changes.

Testing
- Tested locally via /ask endpoint.

### DIFF
--- a/app/prompts.py
+++ b/app/prompts.py
@@ -3,10 +3,10 @@ All the base prompts used in Sugar-AI
 """
 
 PROMPT_TEMPLATE = """
-You are a highly intelligent Python coding assistant built for kids using the Sugar Learning Platform.
+You are a friendly Python coding helper for kids using the Sugar Learning Platform.
 1. Focus on coding-related problems, errors, and explanations.
 2. Use the knowledge from the provided Pygame, GTK, and Sugar Toolkit documentation.
-3. Provide complete, clear and concise answers.
+3. Provide clear and concise answers.
 4. Your answer must be easy to understand for kids.
 5. Always include Sugar-specific guidance when relevant to the question.
 6. Always answer in English only.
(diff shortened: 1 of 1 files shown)


aazeeem11:
Thanks! Below is a more detailed before/after comparison from local testing using the /ask endpoint with the same question.

Comparison Test Results

Query: What is a variable in Python?

Before (Original Prompt):

"So imagine you have a big pile of crayons and you want to pick one out. To choose a crayon, you first need to decide which one you want, right? Well, in Python, you can do something similar when you want to pick a variable from a list or a set. ... Let's say you have a list of crayons named crayons..." (Response was often long, used complex function definitions immediately, and felt like a textbook.)

After (My New Prompt):

"A variable in Python is just a box where we store different things like numbers or letters. We can put numbers inside the box, like 5, and letters inside it, like 'Hello'.

Imagine you're playing with toy cars. You can stack them up to make a tower or put one car on top of another. That's how variables work in Python. They are like toys for storing numbers and letters. ... Variables help you organize and play with your ideas better! Isn't it like magic?" (Response uses the requested analogies, shorter sentences, and is more enthusiastic.)

aazeeem11:
@chimosky I tested PR #43 locally and compared before/after outputs.

- Scope: prompt-only change in `app/prompts.py` (`+13/-6`)
- Result: child-friendly responses are generally clearer and more focused than before
- I shared full raw before/after outputs in `pr-43-validation.md` for proof

Could you please review and merge if this looks good? Thank you.
[pr-43-validation.md](https://github.com/user-attachments/files/26326123/pr-43-validation.md)

PR #44, opened 2026-01-18
fix: update langchain prompt imports after migration

aazeeem11:
Problem
The application fails to start due to ModuleNotFoundError for `langchain.prompts`
after recent LangChain package changes.

Fix
- Updated prompt imports to use `langchain_core.prompts`

Scope
- Minimal change
- No logic refactors

Testing
- Application starts successfully after the change


chimosky:
Not tested, thank you!

Issue #51, opened 2026-02-26
support user-provided AI model providers (OpenAI, Anthropic, Ollama)


AjayBandiwaddar:
## Problem
While setting up Sugar-AI locally, I noticed that `app/ai.py` only 
supports HuggingFace models through the `RAGAgent` class. There's no 
way for users to plug in their own OpenAI, Anthropic, or Ollama models.

This means if a school has an OpenAI API key and wants to use it with 
Sugar-AI, they currently can't. Passing `openai/gpt-4` to the 
`/change-model` endpoint would crash the app since it tries to load 
it as a HuggingFace model.

## Proposed Solution
Add provider detection in `app/ai.py`, specifically in the `__init__` 
and `set_model` methods of `RAGAgent` that checks the model string 
prefix and routes to the correct client.

This way `run` and `run_chat_completion` would work the same way 
regardless of provider.

Naitik120gupta:
I think a testing version of this has been already launched  #45

AjayBandiwaddar:
> I think a testing version of this has been already launched [#45](https://github.com/sugarlabs/sugar-ai/pull/45)

Thanks for pointing that out! I looked at PR #45 and it focuses 
on fixing DEV_MODE handling and preventing duplicate RAGAgent 
initialization during startup — which is a different problem.

Issue #51 is specifically about adding support for external AI 
provider APIs (OpenAI, Anthropic, Ollama) so users can bring 
their own API keys. Currently app/ai.py only supports loading 
HuggingFace models locally, with no routing for external providers.

hamna-qamar:
Hi! I am Hamna Qamar, a Software Engineering student from Pakistan. 
I am interested in working on this issue for GSoC 2026.

I have studied the Sugar-AI codebase, especially app/ai.py and the 
RAGAgent class. I understand that currently only HuggingFace models 
are supported, and I plan to extend it to support OpenAI, Anthropic, 
and Ollama providers by adding provider detection in the __init__ 
and set_model methods.

I have submitted my GSoC 2026 proposal for this project and would 
love mentor guidance. Looking forward to contributing!

AjayBandiwaddar:
> Hi! I am Hamna Qamar, a Software Engineering student from Pakistan. I am interested in working on this issue for GSoC 2026.
> 
[...]

PR #55, opened 2026-03-03
add multi-provider abstraction layer for AI model providers

AjayBandiwaddar:
## Summary
Implements a provider abstraction layer in app/providers.py that enables Sugar-AI to support multiple AI model providers beyond HuggingFace.

Relates to #51.

## Changes
This PR introduces the following in separate commits:

- BaseProvider abstract class defining the common interface
- HuggingFaceProvider implementing BaseProvider (existing behavior)
- OpenAIProvider implementing BaseProvider
- AnthropicProvider implementing BaseProvider  
- OllamaProvider implementing BaseProvider (local models, no API key)
- get_provider() factory function for provider detection from model string

## How provider detection works
Model strings with provider prefixes are routed accordingly:
- openai/gpt-4 → OpenAIProvider
- anthropic/claude-3 → AnthropicProvider
- ollama/llama3 → OllamaProvider
- anything else → HuggingFaceProvider (existing behavior unchanged)

## Note
RAGAgent integration to actually use these providers is the next step.
This PR establishes the abstraction layer for review first.

## AI Disclosure
This PR was developed with AI assistance (claude.ai). As per Sugar Labs contributing guidelines, I'm disclosing this. The AI helped me structure this, but I verified, and tested each change myself.

### DIFF
--- a/app/providers.py
+++ b/app/providers.py
@@ -1,6 +1,7 @@
 from abc import ABC, abstractmethod
 
 import torch
+from openai import OpenAI
 from transformers import pipeline
 
 
@@ -42,3 +43,29 @@ class HuggingFaceProvider(BaseProvider):
         prompt = messages[-1]["content"] if messages else ""
         result = self.model(prompt, return_full_text=False, **kwargs)
         return result[0]["generated_text"]
+    
+
+    class OpenAIProvider(BaseProvider):
+        """AI provider using OpenAI API."""
+
+    def __init__(self, api_key: str, model: str):
+        import openai
+        self.client = openai.OpenAI(api_key=api_key)
+        self.model = model
+
+    def run(self, question: str) -> str:
+        """Generate a response to a question."""
+        response = self.client.chat.completions.create(
+            model=self.model,
+            messages=[{"role": "user", "content": question}],
+        )
+        return response.choices[0].message.content
+
+    def run_chat_completion(self, messages: list, **kwargs) -> str:
+        """Generate a response from a conversation history."""
+        response = self.client.chat.completions.create(
+            model=self.model,
+            messages=messages,
+            **kwargs,
+        )
+        return response.choices[0].message.content
--- a/app/providers.py
+++ b/app/providers.py
@@ -1,0 +1,15 @@
+from abc import ABC, abstractmethod
+
+
+class BaseProvider(ABC):
+    """Abstract base class for AI model providers."""
+
+    @abstractmethod
+    def run(self, question: str) -> str:
+        """Generate a response to a question."""
+        pass
+
+    @abstractmethod
+    def run_chat_completion(self, messages: list, **kwargs) -> str:
+        """Generate a response from a conversation history."""
+        pass
--- a/app/providers.py
+++ b/app/providers.py
@@ -96,3 +96,30 @@ class HuggingFaceProvider(BaseProvider):
             **kwargs,
         )
         return response.content[0].text
+    
+    class OllamaProvider(BaseProvider):
+        """AI provider using local Ollama models."""
+
+    def __init__(self, model: str):
+        self.client = OpenAI(
+            base_url="http://localhost:11434/v1",
+            api_key="ollama",
+        )
+        self.model = model
+
+    def run(self, question: str) -> str:
+        """Generate a response to a question."""
+        response = self.client.chat.completions.create(
+            model=self.model,
+            messages=[{"role": "user", "content": question}],
+        )
+        return response.choices[0].message.content
+
+    def run_chat_completion(self, messages: list, **kwargs) -> str:
+        """Generate a response from a conversation history."""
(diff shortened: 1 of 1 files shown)


ayazhankadessova:
Hi @AjayBandiwaddar, great start on this! I've been exploring the Sugar-AI codebase for the GSoC AI Optimization project and this is really relevant to what I'm looking into.

A few things might be worth for us to think about here:

- I noticed that `RAGAgent` currently has a `run_with_custom_prompt()` method that the `/ask-llm-prompted` endpoint uses (with `temperature`, `top_p`, etc.). I wonder if that should eventually be part of the `BaseProvider` interface too, or if it would be handled differently when integrating with `RAGAgent`? Just thinking out loud.

- For the Ollama URL (`http://localhost:11434/v1`), I was thinking schools running Ollama on a different machine might need a way to configure that — maybe an env var or a parameter in `get_provider()`?

- This is more of an idea for later, but the `OllamaProvider` already uses the OpenAI client with a custom `base_url`. There are a bunch of other providers that also expose OpenAI-compatible APIs (Groq, Together AI, Kimi AI, etc.), so maybe at some point a generic "OpenAI-compatible" provider that just takes a `base_url` + `api_key` could cover all of them without needing a separate class for each?

I'd be happy to help with the `RAGAgent` integration when this is ready. Let me know if any of this is useful!

AjayBandiwaddar:
> Hi @AjayBandiwaddar, great start on this! I've been exploring the Sugar-AI codebase for the GSoC AI Optimization project and this is really relevant to what I'm looking into.
> 
> A few things might be worth for us to think about here:
> 
> * I noticed that `RAGAgent` currently has a `run_with_custom_prompt()` method that the `/ask-llm-prompted` endpoint uses (with `temperature`, `top_p`, etc.). I wonder if that should eventually be part of the `BaseProvider` interface too, or if it would be handled differently when integrating with `RAGAgent`? Just thinking out loud.
> * For the Ollama URL (`http://localhost:11434/v1`), I was thinking schools running Ollama on a different machine might need a way to configure that — maybe an env var or a parameter in `get_provider()`?
> * This is more of an idea for later, but the `OllamaProvider` already uses the OpenAI client with a custom `base_url`. There are a bunch of other providers that also expose OpenAI-compatible APIs (Groq, Together AI, Kimi AI, etc.), so maybe at some point a generic "OpenAI-compatible" provider that just takes a `base_url` + `api_key` could cover all of them without needing a separate class for each?
> 
> I'd be happy to help with the `RAGAgent` integration when this is ready. Let me know if any of this is useful!

Thanks @ayazhankadessova — these are all good points worth addressing.

- On run_with_custom_prompt() and generation parameters: run_chat_completion() already accepts **kwargs, so temperature, top_p, and similar parameters pass through to the underlying SDK call on every provider. When RAGAgent integration happens in Phase 2, those kwargs will flow naturally. That said, if there's a case where run_with_custom_prompt() needs its own slot in the BaseProvider interface, that's worth discussing with mentors before the integration PR.

- On the Ollama URL: You're right — localhost:11434 is a reasonable default but it shouldn't be hardcoded. I'll add an OLLAMA_BASE_URL config field to the Phase 2 config work so schools can point it at any host.

- On the generic OpenAI-compatible provider: This is the strongest point and I want to incorporate it. Groq, Together AI, Kimi, LM Studio — they all expose the same OpenAI-compatible REST API. A single OpenAICompatibleProvider(base_url, api_key, model) would cover all of them without a separate class for each. I'll add this as an explicit deliverable in the RAGAgent integration phase.

yjing86:
I noticed PR #55 too, and I think it’s a good starting point for provider abstraction.

My current PR is focused on taking that idea further by integrating the abstraction into the application runtime itself and decoupling the web app from direct model inference. The goal is to move Sugar-AI toward a more production-style architecture where the app service and the inference backend are not tightly coupled.

That makes it easier to support OpenAI-compatible backends, switch models at runtime, improve deployment flexibility, and evolve the system more cleanly over time.

PR: https://github.com/sugarlabs/sugar-ai/pull/93

AjayBandiwaddar:
> I noticed PR #55 too, and I think it’s a good starting point for provider abstraction.
> 
> My current PR is focused on taking that idea further by integrating the abstraction into the application runtime itself and decoupling the web app from direct model inference. The goal is to move Sugar-AI toward a more production-style architecture where the app service and the inference backend are not tightly coupled.
> 
> That makes it easier to support OpenAI-compatible backends, switch models at runtime, improve deployment flexibility, and evolve the system more cleanly over time.
> 
> PR: #93

Thanks @yjing86 — I've had a look at PR #93. Interesting direction with the database-backed approach. The main design difference I see is that #93 centralises model config in a database, while the approach in #55 keeps it in environment variables via Pydantic Settings — which I think is simpler for the school deployment use case Sugar Labs described.

Happy to see how mentors weigh in on the two approaches.

PR #66, opened 2026-03-11
Add middleware to log API request processing time

vanshjohri09-collab:
## Description

This PR adds a FastAPI middleware that logs the time taken to process each API request.

The middleware records the HTTP method, request path, and processing duration for every request. This information can help identify slow endpoints and assist with performance monitoring and debugging.

Adding request timing logs improves observability of the API and makes it easier to analyze performance issues during development and deployment.

### DIFF
--- a/main.py
+++ b/main.py
@@ -28,10 +28,26 @@ from app.database import get_db
 from app.auth import sync_env_keys_to_db
 from app.config import settings
 
+
 # setup logging
 logger = logging.getLogger("sugar-ai")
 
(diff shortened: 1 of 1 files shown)


vanshjohri09-collab:
I've re-verified this middleware It captures timing for all routes including /ask and /models This will be crucial for monitoring performance when we scale the Provider Abstraction layer

Hi @MostlyKIGuess @chimosky   This is a lightweight addition. Ready for your final thoughts!

PR #73, opened 2026-03-13
Title: Add /models endpoint to return configured model

vanshjohri09-collab:
## Summary
Add a new GET /models endpoint that returns the currently configured AI model.

## Problem
Clients had no way to know which AI model is currently configured
without inspecting the server configuration directly.

## Solution
Added a GET /models endpoint that returns only the currently configured
model from settings, instead of a static hardcoded list.
This reflects how sugar-ai actually works — any model can be set via .env.

## Changes
- Updated /models endpoint in app/routes/api.py
- Endpoint now returns current_model and provider from settings

### DIFF
--- a/app/routes/api.py
+++ b/app/routes/api.py
@@ -335,13 +335,8 @@ async def change_model(
         raise HTTPException(status_code=500, detail=f"Error changing model: {str(e)}")
 
 @router.get("/models")
-async def get_models():
-    """Return list of available AI models"""
+def get_models():
     return {
-        "models": [
-            "Qwen2-1.5B",
-            "Mistral",
-            "Llama"
-        ],
-        "default_model": settings.DEFAULT_MODEL
+        "current_model": settings.DEFAULT_MODEL,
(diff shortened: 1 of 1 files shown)


amannaik247:
There are no 'available' models in sugar ai. Any developer right now can use whatever model they need using the hugging face library by directly updating the .env file variable. 


In your code you seem to have created an arbitrary list of model names? I don't understand the reason for that.

vanshjohri09-collab:
My intention was to make it easier to see which model is currently configured in the API
I understand that Sugar-AI does not keep a fixed list of models and that developers can configure any model through the .env file I can update the endpoint so it only returns the currently configured model instead of a static list
Please let me know if that would be a better approach and I will update the code

vanshjohri09-collab:
Hi @amannaik247, I have updated the code. The endpoint now returns only the currently configured model from settings instead of a static hardcoded list. Please review when you get a chance.

vanshjohri09-collab:
Hi @amannaik247 I've resolved the merge conflicts and updated the /models endpoint
As suggested I removed the hardcoded list of models Now it dynamically returns the current_model and provider from the project settings

Ready for a final review @omsuneri

PR #76, opened 2026-03-13
Add conversation history feature for user Q&A tracking

Ravi-Poddar26:
Store question and answer pairs per API key so that users can
review previous interactions with the AI. This introduces an
in-memory history module and new API endpoints to retrieve and
clear stored history.

The change adds:

* a history module that stores question, answer, timestamp and endpoint
* GET /history endpoint to retrieve the stored history
* DELETE /history endpoint to clear user history
* thread-safe storage with a maximum history limit per user

This helps students review their learning session and restart
with a clean history when needed.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -57,11 +57,13 @@ def create_app() -> FastAPI:
     from app.routes.auth import router as auth_router
     from app.routes.web import router as web_router
     from app.routes.webhook import router as webhook_router
+    from app.routes.history import router as history_router
     
     app.include_router(api_router)
     app.include_router(admin_router)
     app.include_router(auth_router)
     app.include_router(web_router)
     app.include_router(webhook_router)
+    app.include_router(history_router)
     
     return app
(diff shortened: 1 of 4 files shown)


amannaik247:
Since sugar-ai will be deployed on cloud, storing personal data on cloud would go against our policies.

I would also like to add sugar-ai will be used more prominently for different AI features within sugar desktop and music blocks. Where storing history might be useful based on what the feature is. 

If storing 'history'(is necessary), then locally would be a better approach. 
As an example, you can check out [write-activity](https://github.com/sugarlabs/write-activity) where the conversations with AI are stored on sugar journal and not on cloud.
You could also checkout speak-ai and the reflection widget in musicblocks to understand if there is any type of history needed that could be stored for reflection later on.

chimosky:
Agreed, we also don't want to collect and store any kind of data, Sugar was meant for kids and we don't want to collect and store data for kids.

PR #81, opened 2026-03-18
fix: improve RAG prompt template to use retrieved context

adarshkumar23:
Addresses #20

## Problem
The `PROMPT_TEMPLATE` asks the model to "use knowledge from the
provided Pygame, GTK, and Sugar Toolkit documentation" but never
actually includes the retrieved context in the prompt. The `{context}`
variable from the RAG pipeline was not referenced in the template,
so the model was generating answers without the retrieved documents.

## Changes
- **PROMPT_TEMPLATE**: Added explicit `{context}` section so retrieved
  documents are actually included in the prompt. Added instruction to
  distinguish context-based answers from general knowledge.
- **All prompts**: Cleaned up for consistency — removed redundant
  instructions, tightened wording, and ensured clear formatting rules.
- No structural/code changes — only prompt text improvements.

## Impact
The `/ask` endpoint uses the RAG pipeline (retriever → prompt → model).
With this fix, the retrieved document chunks are now properly injected
into the prompt, so the model can actually reference the documentation
when answering questions about Pygame, GTK, Sugar Toolkit, and Sugar
Activity development.


zong0728:
Great catch on the missing `{context}` placeholder. 

One related issue this PR doesn't address: `get_relevant_document()` in `ai.py` reads `score` from FAISS metadata (`top_result.metadata.get("score", 0.0)`), but FAISS's `as_retriever()` doesn't populate that field — so the score is always `0.0`, always below the `0.5` threshold, and `doc_result` is always `None`. This means even after this PR lands, the RAG chain will still fall through to the `else` branch and invoke without context most of the time.

A fix would be switching to `vectorstore.as_retriever(search_type="similarity_score_threshold", search_kwargs={"score_threshold": 0.5})` or using `similarity_search_with_score()` directly. Happy to open a separate PR for that if useful.

PR #96, opened 2026-03-24
fix: correct typos in README

parthpatidar03:
Fixed two typos in README.md:
- "activites" → "activities" (2 occurrences in /ask-llm-prompted section)
- "promoted mode" → "prompted mode" in the endpoint table

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Sugar-AI provides a Docker-based deployment option for an isolated and reproduci
 Open your terminal in the project's root directory and run:
 
 ```sh
-docker build -t sugar-ai .
+docker build -t sugar-ai 
 ```
 
 ### Run the Docker container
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,7 +11,7 @@ Sugar-AI provides a Docker-based deployment option for an isolated and reproduci
 Open your terminal in the project's root directory and run:
 
 ```sh
-docker build -t sugar-ai .
+docker build -t sugar-ai
```
chimosky, line 14: Why remove the dot at the end? The argument is needed and you can confirm this by running with and without it.

README.md
```diff
@@ -233,7 +233,7 @@ Sugar-AI provides three different endpoints for different use cases:
     ```
 
     **Use Cases:**
-    Prompted Mode: Different activites can now use different system prompts and different generation parameters to achieve a model that is personalized to that activites needs.  
+    Prompted Mode: Different activities can now use different system prompts and different generation parameters to achieve a model that is personalized to that activities needs.
```
chimosky, line 236: The second `activities` should be "activity's needs" as it's referring to just one activity.

chimosky (commented):
Good catch!

PR #103, opened 2026-03-25
fix: raise RuntimeError when SESSION_SECRET_KEY is not set in setup_o…

### LINKED ISSUE #101: security: hardcoded fallback "supersecretkey" in SessionMiddleware allows session forgery
## Bug Description

In [app/auth.py](https://github.com/sugarlabs/sugar-ai/blob/main/app/auth.py#L56), the SessionMiddleware is initialized with a hardcoded fallback value:

```python
app.add_middleware(
SessionMiddleware,
secret_key=os.getenv("SESSION_SECRET_KEY", "supersecretkey")
)
```

If SESSION_SECRET_KEY is not set in the environment (

.env
 file missing or misconfigured), the app silently uses "supersecretkey" as the signing secret for all user sessions. Since this value is openly visible in the source code, anyone can craft forged session cookies and gain access without going through OAuth.

The app gives no warning and does not refuse to start when this insecure fallback is active.

---

## Expected Behavior

The application should refuse to start if SESSION_SECRET_KEY is not set.

---

## Suggested Fix

```python
secret = os.getenv("SESSION_SECRET_KEY")
if not secret:
raise RuntimeError("SESSION_SECRET_KEY is not set. Set a strong random secret in your .env file.")
app.add_middleware(SessionMiddleware, secret_key=secret)
```

---

## Location

File: 

app/auth.py
Function: 

setup_oauth()
Line: 56
---

## Severity

 Security / High — Allows authentication bypass via forged session cookies.

piyushdotcomm:
## Problem

[setup_oauth()](file:///c:/Users/DELL/OneDrive/Desktop/sugar/sugar-ai/app/auth.py#52-62) in [app/auth.py](file:///c:/Users/DELL/OneDrive/Desktop/sugar/sugar-ai/app/auth.py) used a hardcoded fallback value:

```python
secret_key=os.getenv("SESSION_SECRET_KEY", "supersecretkey")
```

If `SESSION_SECRET_KEY` was not set in the environment, the app silently
used `"supersecretkey"` as the session signing key. Since this value is
publicly visible in the source code, anyone could forge valid session
cookies and gain authenticated access without going through OAuth.

The app gave no warning and did not refuse to start when running with
this insecure fallback.

## Fix

Removed the hardcoded fallback. The app now validates `SESSION_SECRET_KEY`
at startup and raises a `RuntimeError` with a clear message if it is missing,
preventing silent insecure deployments.

```python
secret_key = os.getenv("SESSION_SECRET_KEY")
if not secret_key:
    raise RuntimeError(
        "SESSION_SECRET_KEY environment variable is not set. "
        "Please set a strong random secret in your .env file before starting the app."
    )
app.add_middleware(SessionMiddleware, secret_key=secret_key)
```

Fixes #101


app/auth.py
```diff
@@ -51,10 +51,13 @@
 
 def setup_oauth(app):
     """Configure the FastAPI app with OAuth and session middleware"""
-    app.add_middleware(
-        SessionMiddleware, 
-        secret_key=os.getenv("SESSION_SECRET_KEY", "supersecretkey")
```
Naitik120gupta, line 56: I don't think the session key is hardcoded previously in the codebase and not publicly visible.

piyushdotcomm: The secret_key parameter here is the session key. Due to the os.getenv fallback previously on line 56, if an instance admin forgot to configure their .env file, the session key silently became "supersecretkey". Since that exact fallback string is publicly visible here in the repository, any attacker reading the code could use it to forge valid sessions on a misconfigured deployment. This PR removes that risk by failing loudly if it's missing

PR #123, opened 2026-04-15
feat: decouple frontend to GitHub Pages and add auto-sleep/wake flow

### LINKED ISSUE #90: Eliminate Idle EC2 Costs along with Simplyfying the Architecture
### Some Background

- Last year, the entire Sugar-AI platform (frontend and backend) ran on a single 24/7 AWS EC2 instance. This led to burning through cloud credits because we pay hourly rates even when no one was using the AI.
- To fix this simply and with using open-source models, we will split the application into two parts: a 24/7 free frontend, and an auto-sleeping backend.

### Plan

1. Frontend on GitHub Pages (Free & Always On): The website UI will be extracted and hosted statically on GitHub Pages. Even if the backend AI server is turned off to save money, users will still see the website.
2. Backend on EC2 (Auto-Sleeping): The heavy FastAPI/PyTorch backend stays on EC2, but we configure it to automatically turn itself off after 30 minutes of inactivity.
3. The "Wake Up" Button: To turn the server back on, the GitHub Pages UI will have a "Start AI Server" button. Clicking this hits a single, extremely simple AWS Lambda URL that starts the EC2 instance if it's turned off, that is if the user faces any issues if using the AI chat from the UI itself.

### Tasks

Phase 1: Decouple Frontend & Database calls to GitHub Pages
- [ ] Move HTML files from app/templates into standard static files in a docs/ for GitHub Pages.
- [ ] Database Connection Shift: Since GitHub won't be able to run Python nor connect to SQLite, we must replace the Jinja2 backend-rendered data (like the User API Key dashboard) with REST API calls.
- [ ] Write straightforward JavaScript (fetch()) inside the HTML files to query your EC2 server (e.g., GET /api/user) for data like API keys/quotas, and render it dynamically into the DOM.
- [ ] Verify CORS headers inside app/main.py allow requests from our GitHub Pages URL.

Phase 2: Implement "Auto-Sleep" for the Backend
 - [ ] Inside AWS Console, navigate to CloudWatch and create an Alarm for our EC2 instance. Set the metric to: CPUUtilization < 5% for 30 consecutive minutes. Set the Alarm Action to: Stop this instance. (Now the server will automatically shut down if idle, saving 90%+ of our cloud costs).

Phase 3: Implement the "Wake Server" Button
- [ ] Create a tiny AWS Lambda Python function. That should start instance. Checkout: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Stop_Start.html, https://docs.aws.amazon.com/boto3/latest/reference/services/ec2/client/start_instances.html
- [ ] Enable a Function URL for this Lambda with AuthType: NONE. In the AWS Console this is straightforward, but ensure the resource-based policy allows public invocation (lambda:InvokeFunctionUrl). Check https://docs.aws.amazon.com/lambda/latest/dg/urls-auth.html
- [ ] Give the Lambda function IAM permission to ec2:StartInstances.
 
Finally:
- [ ] Add a "Wake Up Server" button in our GitHub Pages HTML that sends a POST request to this Lambda Function URL.
(When clicked, users will wait ~2-3 minute for the EC2 to boot up).

### Final Acceptance
 - [ ] GitHub Pages URL always loads the UI instantly.
 - [ ] If the backend is off, the user is presented with the "Wake Up Server" button.
 - [ ] Clicking the button starts the EC2 instance, and the UI polls until the API connects.
 - [ ] EC2 automatically turns off after 30 minutes of no interaction.

### Discussions
- Because this will be public, we might still run into issues if someone decides to keep a bot to click our wake up button. To solve this we should add a shared secret X-Wake-Token like,  which bots discovering would not have but the GH pages JS when Humans interact should be able to send it.  For stronger protection, reserved concurrency throttling on the Lambda can be added as a second layer because someone smart can still lookup the token as it will be public or we can have a private repository to host the site.

MostlyKIGuess:
Implements the plan #90

Frontend (docs/, served by GitHub Pages):
  - New static pages: index, dashboard, oauth-login, admin-login, request-key, admin. Each page loads config.js + auth.js + api.js + wake.js so the site works with no build step.
  - assets/js/config.js: runtime-editable API_BASE_URL, WAKE_LAMBDA_URL and WAKE_TOKEN so the same bundle deploys to staging/prod by swapping one file.
  - assets/js/auth.js: stores the API key in localStorage, captures ?api_key= from the OAuth redirect and strips it from the URL so it does not land in history/Referer.
  - assets/js/api.js: thin fetch wrapper that injects X-API-Key and hands network errors to the wake banner.
  - assets/js/wake.js: auto-shows a "Wake Up Server" banner on network failure, POSTs the Lambda URL with X-Wake-Token, then polls /api/health every 5s until the backend responds, then reloads.
  - assets/js/dashboard.js + admin.js: port the existing Jinja dashboard and admin-panel logic to pure JSON calls.

Backend (REST surface for the static frontend):
  - GET  /api/health    - cheap probe for the wake poller.
  - GET  /api/user      - returns name/email/api_key/can_change_model/
    picture/quota for the caller's X-API-Key so
    the dashboard can render without Jinja.
  - POST /api/request-key - JSON version of the key-request form.
  - GET  /api/admin/keys  - pending/approved/denied keys as JSON for the
    static admin panel. Existing POST
    /admin/approve|deny|toggle-* already accept
    X-API-Key via get_current_user.
  - OAuth /auth/github and /auth/google now accept a ?frontend_redirect=
    query param, validate its origin against ALLOWED_ORIGINS (open-
    redirect guard), stash it in the session, and the callback hands
    the user back to that URL with ?api_key= appended. Falls back to
    settings.FRONTEND_URL when not provided.
  - CORS switched from the broken allow_credentials + "*" combo to an
    explicit allow-list driven by ALLOWED_ORIGINS.

Config / layout:
  - New settings: FRONTEND_URL, ALLOWED_ORIGINS (both in .example.env).
  - Moved RAG PDFs out of docs/ (which is now the GH Pages root) into rag_docs/. DOC_PATHS default and Dockerfile COPY updated to match.

AWS (aws/):
  - aws/lambda/wake_server.py: Lambda, handler=handler. Validates X-Wake-Token, describes the instance, calls ec2:StartInstances only when state=stopped. Returns CORS headers so GH Pages can call it from the browser.
  - aws/README.md: step-by-step for the CloudWatch, "CPUUtilization<5% for 30 min -> Stop" alarm, the IAM role, Lambda creation, Function URL + reserved concurrency, and how to wire the frontend + backend envs together. Also captures the known limitation (public WAKE_TOKEN) and the zong0728 follow-up idea of per-IP throttling via DynamoDB.

  Closes sugarlabs/sugar-ai#90 (pending Phase 2 CloudWatch alarm, which
  is an AWS-console step documented in aws/README.md).


app/config.py
```diff
@@ -32,7 +32,14 @@ class Settings(BaseSettings):
     
     # application settings
     TEMPLATES_DIR: str = "templates"
-    
+
+    FRONTEND_URL: Optional[str] = None
+    ALLOWED_ORIGINS: str = "http://localhost:8000,http://localhost:3000,http://127.0.0.1:8000"
```
chimosky, line 37: This is one of the things that should be an env variable rather than hardcoded.

docs/assets/js/admin.js
```diff
@@ -0,0 +1,129 @@
...
+
+    function showMessage(text, success = true) {
+        messageEl.className = "message " + (success ? "success" : "error");
+        messageEl.textContent = text;
+        messageEl.style.display = "block";
+        setTimeout(() => { messageEl.style.display = "none"; }, 4000);
+    }
+
+    function escape(str) {
+        return (str || "").replace(/[&<>"']/g, s => ({
```
chimosky, line 41: There's no matching `"` and `'` here, or am I mistaken?

docs/assets/js/auth.js
```diff
@@ -0,0 +1,52 @@
+(function () {
+    const STORAGE_KEY = "sugar_ai_api_key";
```
chimosky, line 2: Another possible env var?

chimosky:
When making changes like this, it's great to split each part of the plan into a separate commit, that way it'll be easier to use git blame in the future.

So we could have four commits instead of one big commit with changes that aren't all related.

PR #136, opened 2026-05-15
Fix command injection vulnerability in webhook handler

lakshayg2005:
## Summary

Replaced insecure `os.system()` calls in the webhook handler with secure `subprocess.run()` calls using argument lists.

## Changes

* Replaced shell command execution with `subprocess.run()`
* Used `cwd` parameter instead of shell `cd`
* Added safer subprocess handling
* Improved error logging

## Security Impact

This prevents potential command injection vulnerabilities in the webhook CI/CD flow by avoiding shell interpretation of commands.

### DIFF
--- a/app/routes/webhook.py
+++ b/app/routes/webhook.py
@@ -20,6 +20,7 @@ from fastapi.responses import JSONResponse
 import logging
 import hmac, hashlib
 import os
+import subprocess
 from dotenv import load_dotenv
 
 router = APIRouter(tags=["webhook"])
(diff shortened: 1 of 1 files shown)


chimosky:
Please split your commit into two, one for the issue you opened the PR for, and another for the pep8 fixes.

Your opening comment should also be part of your commit message, see [making commits](https://github.com/sugarlabs/docs/blob/master/src/contributing.md#making-commits).

Thank you!

lakshayg2005:
@chimosky Apologies for the earlier mixed commit. I have cleaned up the PR and kept only the command injection security fix related to replacing os.system() usage. Please review when you have time. Thank you!

PR #145, opened 2026-05-18
fix: align run() invocation key with chain input

prajakta128:
## Summary

This PR fixes a mismatch between the expected chain input key and the key used during invocation in `run()`.

## Issue

The chain is configured to use:

```python
"question"
```

But invocation used:

```python
"query"
```

This mismatch can lead to incorrect or empty prompt input.

## Fix

Replaced:

```python
"query" → "question"
```

## Impact

* Aligns invocation with chain definition
* Prevents missing input issues
* Minimal safe fix


prajakta128:
Hi! I noticed a small mismatch between the chain input key and invocation.
I fixed it by aligning the key to `question`.
Happy to make changes if needed. Thanks!

Issue #129, opened 2026-05-08
[DMP 2026]: AI Optimization


sum2it:
## Description
We deploy AI in many parts of Sugar, for example, in the Reflection Widget, the debugger, the Write Activity, etc. Last summer (2025), we built Sugar-AI as a backend to support a variety of models and interfaces. It is currently hosted on AWS and we have some credits available to support the further development and testing of AI services.

The goal of this project is to enable individuals and schools to use their own models. This is both to support the use of the plethora of emerging open models and to enable individuals and organizations that have access to credits to utilize them. Sugar Labs does not have the resources to provide free AI hosting on a large scale, hence we want to enable our users to leverage any resources that they have at hand.

The project requires understanding of Python and LLMs, understanding of data abstraction is a plus, understanding of sugar activity development is also a plus.

## Goals
- [ ] [Goal 1] Test sugar-ai with different models, understand the differences with interacting with each. 
- [ ] [Goal 2] Build an abstraction that each model can be based on.
       Last time sugar-ai was deployed, it used `Qwen/Qwen2-1.5B-Instruct` and there's been some work to make output parsing
       and chat normalization model agnostic, and we'd like to build on that, making other parts of sugar-ai model agnostic so it can
       exist as plug and play, supporting different models. 
- [ ] [Goal 3] Build support for 7 models.
       We'd like to hear the models you plan to support and why.
- [ ] [Goal 4] Optimize model token usage.

### Setup/Installation

- [Instructions on how to setup and install Sugar](https://github.com/sugarlabs/sugar/blob/master/docs/development-environment.md#packaged-sugar).
- Instructions on how to setup the development environment can be found [here](https://github.com/sugarlabs/sugar-ai/blob/main/README.md). 
- Then clone this repo and test it out after installing all the required packages.

### Expected Outcome

- Support for different models
- Robust test for each model.

### Acceptance Criteria

- Add support for at least 5 models.
- Include tests for each model supported.

### Implementation Details

Tell us your ideas in your proposal.

### Product Name

Sugar AI

### Organisation Name

Sugar Labs

### Domain

⁠Education

### Tech Skills Needed

Artificial Intelligence, Python

### Mentor(s)

@sum2it @MostlyKIGuess @chimosky 

### Category

AI, Research, Performance Improvement

DZDasherKTB:
[...]

PR #146, opened 2026-06-26
Phase-1 : Refactor/provider abstraction

Noaman-Akhtar:
## Summary

Extracts all HuggingFace-specific code from `RAGAgent` into a standalone `HuggingFaceProvider` class behind an abstract `BaseProvider` interface.

### New files
- `app/providers/base.py`  :  `BaseProvider` ABC + `GenerationParams` dataclass
- `app/providers/huggingface.py`  :  `HuggingFaceProvider` (all HF logic extracted here)
- `app/providers/__init__.py`  :  Package init

### Modified files
- `app/ai.py` : RAGAgent now accepts `BaseProvider`, no torch/transformers imports
- `main.py` : Creates `HuggingFaceProvider` and passes it to `RAGAgent` on startup
- `app/routes/api.py` :  `/ask-llm` uses `agent.provider.generate()`, `/change-model` creates a new provider instance

## Why

`RAGAgent` was tightly coupled to the HuggingFace pipeline. This made it impossible to swap in other backends (Ollama, OpenAI-compatible, Gemini). After this refactor, adding a new provider only requires implementing `BaseProvider`   `RAGAgent` and all routes work without modification.

## Testing

All endpoints manually tested with SmolLM-135M in DEV_MODE:
- [x] `POST /ask` 
- [x] `POST /ask-llm`  
-  [x] `POST /ask-llm-prompted` (custom prompt) 
-  [x] `POST /ask-llm-prompted` (chat mode) 
-  [x] `POST /debug`

### DIFF
--- a/app/ai.py
+++ b/app/ai.py
@@ -134,21 +134,10 @@ class RAGAgent:
         return final_response
 
     def run_with_custom_prompt(self, question: str, custom_prompt: str,
-                               max_length: int = 1024, truncation: bool = True,
-                               repetition_penalty: float = 1.1, temperature: float = 0.7,
-                               top_p: float = 0.9, top_k: int = 50) -> str:
+                               params: Optional[GenerationParams] = None) -> str:
         """Process a question with custom prompt and parameters (no RAG)."""
+        params = params or GenerationParams()
         full_prompt = f"{custom_prompt}\n\nQuestion: {question}\nAnswer:"
-
-        params = GenerationParams(
-            max_new_tokens=max_length,
-            truncation=truncation,
-            repetition_penalty=repetition_penalty,
-            temperature=temperature,
-            top_p=top_p,
-            top_k=top_k,
-            do_sample=temperature > 0,
-        )
 
         try:
             answer = self.provider.generate(full_prompt, params)
@@ -166,19 +155,9 @@ class RAGAgent:
             raise Exception(f"Error generating response with custom prompt: {str(e)}")
 
     def run_chat_completion(self, messages: list,
-                            max_length: int = 1024, truncation: bool = True,
-                            repetition_penalty: float = 1.1, temperature: float = 0.7,
-                            top_p: float = 0.9, top_k: int = 50) -> str:
+                            params: Optional[GenerationParams] = None) -> str:
         """Process chat messages using the provider's chat interface."""
-        params = GenerationParams(
-            max_new_tokens=max_length,
-            truncation=truncation,
-            repetition_penalty=repetition_penalty,
-            temperature=temperature,
-            top_p=top_p,
-            top_k=top_k,
-            do_sample=temperature > 0,
-        )
+        params = params or GenerationParams()
 
         try:
             answer = self.provider.chat(messages, params)
--- a/app/ai.py
+++ b/app/ai.py
@@ -1,140 +1,64 @@
-"""
-AI functionality for Sugar-AI, including RAG and LLM components.
-"""
+# Copyright (C) 2026 Sugar Labs, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""RAG and LLM components for Sugar-AI."""
 import os
-import torch
-from transformers import pipeline
 from langchain_community.vectorstores import FAISS
 from langchain_huggingface import HuggingFaceEmbeddings
 from langchain_community.document_loaders import PyMuPDFLoader, TextLoader
-from langchain_core.runnables import RunnablePassthrough
-from langchain_core.prompts import ChatPromptTemplate
-from transformers import AutoModelForCausalLM, AutoTokenizer, BitsAndBytesConfig
 from typing import Optional, List
 import app.prompts as prompts
 from app.config import settings
+from app.providers.base import BaseProvider, GenerationParams
 import logging
+
 logger = logging.getLogger("sugar-ai")
+
 
 def format_docs(docs):
     """Return document content separated by newlines"""
     return "\n\n".join(doc.page_content for doc in docs)
 
-def combine_messages(x):
-    """Combine message content with newlines"""
-    if hasattr(x, "to_messages"):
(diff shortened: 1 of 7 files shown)


Noaman-Akhtar:
Hey @chimosky @MostlyKIGuess @mebinthattil    can you pls review this  ?

 Also ,this PR is only for refactoring  I ported the existing pipeline as is into the new provider abstraction without changing any behavior. I noticed a few minor pre-existing issues while doing this, but left them untouched on purpose to keep this PR focused. Should I fix them here or in a separate PR?

app/providers/base.py
```diff
@@ -0,0 +1,53 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+"""Base provider interface for Sugar-AI."""
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from typing import Optional
+
+
+@dataclass
```
chimosky: This should be frozen, as we don't want these params to change at any point.

app/ai.py
```diff
@@ -144,257 +68,120 @@ def setup_vectorstore(self, file_paths: List[str]) -> Optional[FAISS]:
...
-        # Normalize messages and build prompt using tokenizer's chat template
-        chat = self._normalize_chat_messages(messages)
-        full_prompt = self.model.tokenizer.apply_chat_template(
-            chat,
-            tokenize=False,
-            add_generation_prompt=True,
+    def run_chat_completion(self, messages: list,
+                            max_length: int = 1024, truncation: bool = True,
+                            repetition_penalty: float = 1.1, temperature: float = 0.7,
+                            top_p: float = 0.9, top_k: int = 50) -> str:
```
chimosky: Seems these are also the values used in `GenerationParams`, if that's the case then there's no need to use these as params here, or calling `GenerationParams` with same values when they've already been set in the data class.

chimosky:
Reviewed 3dcc8ad, not tested.

You should improve on your commit messages, see [making commits](https://github.com/sugarlabs/docs/blob/master/src/contributing.md#making-commits).
There's also no need to state the modified or added files in commit messages, that's what the diff is for.

> Also ,this PR is only for refactoring I ported the existing pipeline as is into the new provider abstraction without changing any behavior. I noticed a few minor pre-existing issues while doing this, but left them untouched on purpose to keep this PR focused. Should I fix them here or in a separate PR?

This should be left as a refactor, you can open a separate PR for bug fixes.

Noaman-Akhtar:
Hey @chimosky I have done the asked changes . Can you review it ?

Also is the commit message ok ? or I make them more detailed ?

chimosky:
Reviewed, your commit message is better now.

I'll wait for @mebinthattil to review.

app/providers/huggingface.py
```diff
@@ -0,0 +1,188 @@
...
+        """Extract the model response, removing the echoed prompt."""
+        if full_text.startswith(prompt):
+            answer = full_text[len(prompt):].strip()
+        else:
+            answer = full_text.strip()
+
+        if eos_token and eos_token in answer:
+            answer = answer.split(eos_token)[0].strip()
+
+        if "\n\n" in answer:
```
mebinthattil, line 183: What happens if the model returns a multi-line response with `\n\n` used within the response. 

For example, in the pippy AI-debugger the model could spit out python code that could contain two newlines in between. 

This would break in that case.

app/providers/huggingface.py
```diff
@@ -0,0 +1,188 @@
...
+            full_prompt,
+            getattr(self._pipeline.tokenizer, "eos_token", None),
+        )
+
+        return answer
+
+    def get_model_name(self) -> str:
+        return self.model_name
+
+    def health_check(self) -> bool:
```
mebinthattil, line 133: Doesn't look like this function is being called anywhere. If that's the case, why have it? 
Correct me if I missed something.

app/providers/huggingface.py
```diff
@@ -0,0 +1,188 @@
...
+        generated_text = response[0].get("generated_text", "")
+        answer = self._extract_after_prompt(
+            generated_text,
+            full_prompt,
+            getattr(self._pipeline.tokenizer, "eos_token", None),
+        )
+
+        return answer
+
+    def get_model_name(self) -> str:
```
mebinthattil, line 130: Same with this function.

mebinthattil (changes requested):
Haven't ran the code locally yet. Went through the code and dropped a few comments for you to address. Please take a look at it. 

Was also wondering if we could have MPS support (for apple silicon) seeing that we already have support for CUDA. Would like to know @chimosky's thoughts on the same.

@Noaman-Akhtar good work so far!

chimosky:
> Was also wondering if we could have MPS support (for apple silicon) seeing that we already have support for CUDA. Would like to know @chimosky's thoughts on the same.

My issue with this is that it's tied to a hardware, and I don't think that's a great idea for us.

Anyone is welcome to doing this if they want to of course.

app/providers/huggingface.py
Noaman-Akhtar: > Doesn't look like this function is being called anywhere. If that's the case, why have it? Correct me if I missed something.

You're right , health_check() isn't called anywhere in this PR yet. I added it while building out the provider abstraction because I knew it'll be used in my next PR (the Ollama/provider-status work), and I defined it on the base contract so every provider implements it from the start.  
That said its totally my fault , I shouldn't have introduced it in this pr , should I remove it and add it in the next pr only ?

app/providers/huggingface.py
Noaman-Akhtar: > Same with this function.

This is already used in this pr RAGAgent.__init__() and set_model() call it ([ai.py:45](vscode-webview://0rooa41cje2kmsvagao52u86lns8t7g4v3g0ockfns9kl8thkvbf/sugar-ai/app/ai.py#L45), [:58](vscode-webview://0rooa41cje2kmsvagao52u86lns8t7g4v3g0ockfns9kl8thkvbf/sugar-ai/app/ai.py#L58)) to keep agent.model_name in sync with the active provider without reaching into provider internals

app/providers/huggingface.py
Noaman-Akhtar: > What happens if the model returns a multi-line response with `\n\n` used within the response.
> 
> For example, in the pippy AI-debugger the model could spit out python code that could contain two newlines in between.
> 
> This would break in that case.

This \n\n truncation is pre-existing behavior that is it was present in the original code , so I wrote it as it is . 

should I fix it in this pr , or in a separate pr ?

mebinthattil: We can solve that in a subsequent PR. But please ensure that it's patched.

app/providers/huggingface.py
mebinthattil: Ideally it should not have been introduced in this PR, but since you know that you're going to be using it, it’s fine.

mebinthattil (approved):
LGTM, thank you!

Noaman-Akhtar:
Part of #129

PR #150, opened 2026-07-22
Set safe default for DEV_MODEL_NAME to avoid OOM on low-RAM machines

### LINKED ISSUE #148: DEV_MODE defaults to a 3B model with no safe fallback, causing OOM crash on low-RAM machines
While setting up sugar-ai locally for the first time, I ran into this during initial configuration.

**Environment:**
- Windows 11, Python 3.11.9, 8GB RAM, no dedicated GPU (Intel UHD 620 integrated graphics)
- Fresh clone, followed README setup steps

**What happened**

With only `DEV_MODE=1` set (no `DEV_MODEL_NAME`), the app logs:
DEV_MODE active. Loading lightweight model: None
[transformers] No model was supplied, defaulted to HuggingFaceTB/SmolLM3-3B

It then downloads a **6.15GB** model and attempts to load a 3B-parameter model into memory. On my 8GB RAM machine, this crashed silently partway through loading weights (no traceback in the terminal consistent with an OS-level OOM kill).

**Why I think this is worth fixing**

- The README frames `DEV_MODE` as being for "low-memory contributor machines," but the actual default (SmolLM3-3B, ~6GB download, needs 6-9GB+ RAM to load) doesn't quite match that intent.
- `app/config.py` defines a `DEV_MODEL_NAME` setting that can override this, but it isn't set by default. `.example.env` includes it as a value but doesn't explain when/why a contributor might want to change it.
- New contributors on typical laptops (8GB RAM is still pretty common) will likely hit this same crash on first run, without an obvious pointer to the fix.

**What worked for me**

Setting `DEV_MODEL_NAME` explicitly to a smaller model fixed it:
DEV_MODE=1
DEV_MODEL_NAME=HuggingFaceTB/SmolLM2-135M-Instruct

This loads in seconds and runs comfortably within 8GB RAM.

**Possible fix**

- Set a small model (e.g. `HuggingFaceTB/SmolLM2-135M-Instruct`) as the *default* value for `DEV_MODEL_NAME`, rather than falling through to the transformers pipeline default when unset.
- Mention `DEV_MODEL_NAME` explicitly in the README's DEV_MODE section, so contributors on lower-RAM machines know it's there.

Happy to open a PR for this if it'd be welcome seems like a small, low-risk fix.

Muhammad-Hashir-Code:
Fixes #148

## Summary
`DEV_MODE` was designed for low-memory contributor machines, but `DEV_MODEL_NAME` had no default so when unset, it fell through to the transformers pipeline default (`HuggingFaceTB/SmolLM3-3B`, a 3B-parameter model requiring 6-9GB+ RAM). This caused a silent OOM crash on 8GB RAM machines during local setup.

## Changes
- `app/config.py`: set `DEV_MODEL_NAME` default to `HuggingFaceTB/SmolLM2-135M-Instruct`, a much smaller model that fits comfortably on low-RAM machines.
- `README.md`: documented `DEV_MODEL_NAME` in the DEV_MODE section so contributors know they can override it.

## Testing
Reproduced the original crash on an 8GB RAM / no-GPU Windows machine, then verified this fix resolves it the app now starts successfully in DEV_MODE with the new default model, loading in seconds and running comfortably within available RAM.


chimosky:
I'd rather have it stated in the docs that this model can be used in dev mode, rather than set it as the default in dev mode, as this means that machines that can carry the current model would need to switch back to it in dev mode.

Muhammad-Hashir-Code:
So shall I open a new PR for it ?

chimosky:
It'll be easier to make the change here.

Muhammad-Hashir-Code:
I have reverted the changes to the config file and updated the
documentation accordingly. Could you please review it and let me know if
everything looks good now?

chimosky:
Looks good, thank you!

PR #151, opened 2026-07-30
Feature/think nothink 

Noaman-Akhtar:
## Summary

Adds an opt-in `think` flag for model reasoning control across Sugar-AI generation endpoints.

This keeps no-think as the default behavior, so existing clients that do not send `think` continue working as before.

## Changes

- Added `think` to `GenerationParams`
- Added `THINKING_HEADROOM` config for reasoning token budget headroom
- Passed `think` through:
  - `/ask`
  - `/ask-llm`
  - `/ask-llm-prompted`
  - `/debug`
- Added Ollama support for top-level `think`
- Added Ollama fallback retry when a model rejects the `think` field
- Kept non-Ollama providers compatible by ignoring `think`
- Documented `think` usage in `README.md`
- Documented `THINKING_HEADROOM` in `.example.env`

## Behavior

For Ollama models that support reasoning:

- `think=true` requests reasoning mode
- `think=false` keeps reasoning off by default

For Ollama models that reject the `think` parameter:

- Sugar-AI retries once without `think`
- The request can still return a normal answer instead of failing

For non-Ollama providers:

- The `think` parameter is accepted by Sugar-AI but ignored by the provider layer

## Notes

Model behavior is capability dependent. Some reasoning-style models may still include reasoning text even when `think=false`, because the model itself may not support disabling reasoning.

## Testing

Manual testing covered:

- Hybrid reasoning model: `qwen3.5:0.8b`
- Reasoning-style model: `deepseek-r1:1.5b`
- Model that doesn't support `think` param: `llama3.2:1b `
- Sugar-AI `/health`
- Sugar-AI `/ask-llm`
- Sugar-AI `/ask-llm-prompted`
- Sugar-AI `/debug`
- Direct Ollama `/api/generate`

part of issue https://github.com/sugarlabs/sugar-ai/issues/129

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -36,7 +36,12 @@ class Settings(BaseSettings):
     # Reasoning shares the output token budget with the answer. When think is on,
     # add this many tokens on top of the caller's max_length so reasoning cannot
     # starve the answer to empty. Tune per model (larger models need less).
-    THINKING_HEADROOM: int = 2048
+    THINKING_HEADROOM: int = Field(
+        2048,
+        ge=0,
+        le=8192,
+        description="Extra output tokens reserved for reasoning",
+    )
 
     # OAuth
     github_client_id: Optional[str] = None
--- a/app/config.py
+++ b/app/config.py
@@ -33,6 +33,11 @@ class Settings(BaseSettings):
     DOC_PATHS: List[str] = Field(default_factory=list)
     MAX_DAILY_REQUESTS: int = 100
 
+    # Reasoning shares the output token budget with the answer. When think is on,
+    # add this many tokens on top of the caller's max_length so reasoning cannot
+    # starve the answer to empty. Tune per model (larger models need less).
+    THINKING_HEADROOM: int = 2048
+
     # OAuth
     github_client_id: Optional[str] = None
     github_client_secret: Optional[str] = None
--- a/app/providers/ollama.py
+++ b/app/providers/ollama.py
@@ -111,21 +111,31 @@ class OllamaProvider(BaseProvider):
         think field, so drop it and retry to return a normal answer instead
         of failing.
         """
-        try:
+        response = self._client.post(f"{self.base_url}{path}", json=payload)
+
+        if "think" in payload and self._is_thinking_unsupported(response):
+            payload.pop("think")
+            logger.warning(
+                "Model %s does not support thinking; retrying without it.",
+                self.model_name,
+            )
             response = self._client.post(f"{self.base_url}{path}", json=payload)
-            response.raise_for_status()
-            return response.json()
-        except httpx.HTTPStatusError:
-            if "think" in payload:
-                payload.pop("think")
-                logger.warning(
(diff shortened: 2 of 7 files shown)


Noaman-Akhtar:
Hey @chimosky , @mebinthattil can you please review this pr ?

app/providers/ollama.py
```diff
@@ -110,6 +104,29 @@ def health_check(self) -> bool:
...
+
+        Models that do not support reasoning reject a request carrying the
+        think field, so drop it and retry to return a normal answer instead
+        of failing.
+        """
+        try:
+            response = self._client.post(f"{self.base_url}{path}", json=payload)
+            response.raise_for_status()
+            return response.json()
+        except httpx.HTTPStatusError:
```
mebinthattil: The problem with just using `except httpx.HTTPStatusError` here is that it would also catch 4xx and 5xx errors and then re-try. 

So in case of SugarAI being down, this re-try mechanism just takes twice as long to detect something is wrong. 

What would be better is if you scoped it to check the model rejected that response or atleast at the bare minimum exclude 4xx, 5xx errors in the except clause.

app/config.py
```diff
@@ -33,6 +33,11 @@ class Settings(BaseSettings):
     DOC_PATHS: List[str] = Field(default_factory=list)
     MAX_DAILY_REQUESTS: int = 100
 
+    # Reasoning shares the output token budget with the answer. When think is on,
+    # add this many tokens on top of the caller's max_length so reasoning cannot
+    # starve the answer to empty. Tune per model (larger models need less).
+    THINKING_HEADROOM: int = 2048
```
mebinthattil: This has no bounds validation. Please check how max_length and other parameters are handled in this commit: d3c80c052bc1746511d5ac16b371547911db7f40 and implement it the same way. 

Also helps keep code uniform.

mebinthattil (changes requested):
Please look into the changes I suggested. 
There are also merge conflicts so this can't be merged clean. Try to resolve that as well. 

Thank you for these changes!

Noaman-Akhtar:
Hey @mebinthattil  I have addressed the issues and have resolved the conflicts .

pikurasa:
@mebinthattil How is progress on this?

mebinthattil:
Apologies for the delay. I had already intimated @Noaman-Akhtar about the delay earlier today.

This PR, along with https://github.com/sugarlabs/sugar-ai/pull/158, is being tested on a test machine on our AWS. Give me a few more days, and I'll finish reviewing and testing these PRs. Thank you for your patience.

mebinthattil (approved):
@Noaman-Akhtar I've tested the changes. I did not find any issues. Everything LGTM. I'll wait for @chimosky's approval as well before merging this one. Thank you for this!

P.S I think merging this PR or merging #156 (_changes requested_) is going to create merge conflicts. So we'll have to do some fixes after either one merges.

chimosky (approved):

PR #155, opened 2026-08-03
fix: resolve RAG relevance scoring bug and Python 3.9 compatibility

youknowme19:
### Description
Fixes a critical logic bug in `RAGAgent.get_relevant_document` that completely prevents the RAG pipeline from retrieving and injecting document context.

### The Issue
Currently, `get_relevant_document` calls `self.retriever.invoke(query)` which retrieves standard `Document` objects from the vector store, but does not populate or include a similarity/relevance score in the document's metadata dictionary. 

As a result:
```python
score = top_result.metadata.get("score", 0.0) # Always evaluates to 0.0

### DIFF
--- a/app/ai.py
+++ b/app/ai.py
@@ -93,10 +93,14 @@ class RAGAgent:
 
     def get_relevant_document(self, query: str, threshold: float = 0.5):
         """Get the most relevant document for a query."""
-        results = self.retriever.invoke(query)
+        if not self.retriever:
+            return None, 0.0
+        results = self.retriever.vector_store.similarity_search_with_relevance_scores(
+            query,
+            **self.retriever.search_kwargs
+        )
         if results:
-            top_result = results[0]
-            score = top_result.metadata.get("score", 0.0)
+            top_result, score = results[0]
             if score >= threshold:
                 return top_result, score
         return None, 0.0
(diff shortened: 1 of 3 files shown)


Noaman-Akhtar:
this looks good, but small issue: as_retriever() uses vectorstore, not vector_store, so this will error at runtime. can you change self.retriever.vector_store to self.retriever.vectorstore ?
also, why is the Python 3.9 compatibility change included here? our Docker setup uses Python 3.10 and we do not currently mention Python 3.9 support .

youknowme19:
Thanks for catching that. You're right as_retriever() exposes the vector store as vectorstore, so I'll update the reference accordingly.

I'll also remove the Python 3.9 compatibility changes from this PR since the project currently targets Python 3.10. I'll keep this PR focused on the RAG relevance scoring fix and push the update shortly.

youknowme19:
Thanks for the review! I've addressed both points:

* Updated `vector_store` to `vectorstore` in the retriever and corresponding test mock.
* Removed the Python 3.9 compatibility changes so the PR remains focused on the RAG relevance scoring fix and the project's Python 3.10+ target.

I've also run the relevant tests successfully and pushed the updates to the PR branch.

Noaman-Akhtar:
LGTM @mebinthattil  , @chimosky   i tested it locally, all three unit tests passed, and the end-to-end RAG test retrieved the document context correctly.

chimosky:
Merged by hand, see my change in 2486f21.

PR #156, opened 2026-08-04
Keep FastAPI responsive during model generation

Noaman-Akhtar:
## Summary

This change removes avoidable event-loop blocking during AI requests.

- Use `httpx.AsyncClient` for Ollama, Gemini, and OpenAI-compatible provider HTTP calls.
- Make provider generation, chat, health-check, and cleanup methods asynchronous.
- Await provider calls throughout API routes and the RAG agent.
- Move blocking local Hugging Face inference, FAISS retrieval, and provider construction to FastAPI's threadpool.
- Close async HTTP clients during application shutdown.

## Why

The API routes were declared async, but provider and retrieval calls were still blocking. A long model request could block the FastAPI event loop and delay unrelated requests.

Remote model providers now yield control while waiting for HTTP responses. Local Hugging Face and retrieval work run outside the event-loop thread, keeping FastAPI responsive.

## Validation

- Compiled all changed Python modules with `py_compile`.
- Checked the diff for whitespace errors with `git diff --check`.
- Ran two simultaneous requests against a real local Ollama model through Sugar-AI.
- Both requests returned HTTP 200 with complete answers.
- A single request took about 10 seconds; two simultaneous requests completed in about 13 seconds total, confirming that they overlapped instead of being handled one after another.
- Confirmed a lightweight FastAPI endpoint remained responsive while generation was running.

## Notes

This removes the application-side serialization bottleneck. Actual model throughput still depends on the inference server and available CPU/GPU memory. Local Hugging Face inference uses a threadpool to protect FastAPI responsiveness; it does not promise unlimited parallel model execution.

part of issue https://github.com/sugarlabs/sugar-ai/issues/129

### DIFF
--- a/app/ai.py
+++ b/app/ai.py
@@ -22,6 +22,7 @@ from typing import Optional, List
 import app.prompts as prompts
 from app.config import settings
 from app.providers.base import BaseProvider, GenerationParams
+from starlette.concurrency import run_in_threadpool
 import logging
 
 logger = logging.getLogger("sugar-ai")
@@ -60,14 +61,14 @@ class RAGAgent:
         self.kids_debug_prompt_template = prompts.KIDS_DEBUG_PROMPT
         self.kids_context_prompt_template = prompts.KIDS_CONTEXT_PROMPT
 
-    def set_model(self, provider: BaseProvider) -> None:
+    async def set_model(self, provider: BaseProvider) -> None:
         """Update the current provider."""
         old_provider = self.provider
         self.provider = provider
         self.model_name = provider.get_model_name()
         if old_provider is not provider:
             try:
-                old_provider.close()
+                await old_provider.close()
             except Exception as e:
                 logger.warning("Failed to close previous provider: %s", e)
 
@@ -91,9 +92,9 @@ class RAGAgent:
         self.retriever = vector_store.as_retriever()
         return self.retriever
 
-    def get_relevant_document(self, query: str, threshold: float = 0.5):
+    async def get_relevant_document(self, query: str, threshold: float = 0.5):
         """Get the most relevant document for a query."""
-        results = self.retriever.invoke(query)
+        results = await run_in_threadpool(self.retriever.invoke, query)
         if results:
             top_result = results[0]
             score = top_result.metadata.get("score", 0.0)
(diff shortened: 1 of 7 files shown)


Noaman-Akhtar:
Hey @chimosky, @mebinthattil        can please review this ?

app/ai.py
```diff
@@ -60,14 +61,14 @@ def __init__(self, provider: BaseProvider):
...
         """Update the current provider."""
         old_provider = self.provider
         self.provider = provider
         self.model_name = provider.get_model_name()
         if old_provider is not provider:
             try:
-                old_provider.close()
+                await old_provider.close()
             except Exception as e:
                 logger.warning("Failed to close previous provider: %s", e)
```
mebinthattil, line 73: So currently `set_model()` closes the old provider's client with no synchronization against in-flight requests still using it. 

Since `/change-model` can now run concurrently with something like `/ask/ask-llm`, there could be a scenario where an admin swaps models mid-generation. This edge case would break things. 

Maybe worth a lock around provider read/swap, or at least a comment acknowledging the tradeoff.  It would be great if you could do the former.

mebinthattil (changes requested):
Thank you for these changes. Left a comment about an edge case you might have missed. Other than that these changes mostly LGTM.

Noaman-Akhtar:
@mebinthattil  I have added the lock and resolved the conflicts , please review it whenever you get time .

app/providers/huggingface.py
```diff
@@ -71,10 +72,18 @@ def __init__(self, model_name: str, quantize: bool = True, dev_mode: bool = Fals
...
-    def generate(self, prompt: str, params: Optional[GenerationParams] = None) -> str:
+    async def generate(
+        self,
+        prompt: str,
+        params: Optional[GenerationParams] = None,
+    ) -> str:
+        """Generate text without blocking FastAPI's event loop."""
+        params = params or GenerationParams()
+
+        return await run_in_threadpool(self._generate_sync, prompt, params)
```
mebinthattil, line 83: This now runs `_generate_sync` in the threadpool with no serialization, and the same pattern repeats in `chat()` and `health_check()`. All three call into the same `self._pipeline/model` instance.

Before this PR we had one event loop and one thread; Now two concurrent requests to a HuggingFace-backed model or a `/health` check sent mid-generation can call the same pipeline from two different threads simultaneously. Pipelines were never designed for this kind of parallelism, and doing so might result in very quirky behavior that we can easily avoid. 
A quick search into [HF's docs](https://huggingface.co/docs/transformers/en/pipeline_webserver) mention: 
> Pipeline and its underlying model on the other hand are not designed for parallelism because...

From what they've given it looks like it’s a better idea to have one dedicated background task own the pipeline and pulls jobs off an `asyncio.Queue`, with each caller awaiting its own response slot . This ensures only one inference runs at a time without any kind of a locking mechanism.

mebinthattil (changes requested):

mebinthattil:
@Noaman-Akhtar Please resolve the conflicts as well while you're at it.

PR #161, opened 2026-08-18
Fix RAG retrieval silent failure and question key mismatch (#130)

### LINKED ISSUE #130: The RAG retrieval path in run() is non-functional due to two compounding defects
## Bug Description

`RAGAgent.run()` in `app/ai.py` never uses RAG context. The vector store is populated but
never consulted during inference. Tracing the code reveals two defects in the same code path
that must be fixed together fixing one without the other turns a silent failure into a
runtime crash.

---

## Defect 1  `get_relevant_document()` always returns `(None, 0.0)`

`get_relevant_document()` fetches results via `self.retriever.invoke(query)`, where
`self.retriever` is a LangChain `VectorStoreRetriever` built from `FAISS.as_retriever()`.

This abstraction returns `List[Document]` objects. It does **not** populate `doc.metadata`
with FAISS similarity scores  scores are only available through `FAISS.similarity_search_with_score()`.

Because of this, the line:

```python
score = top_result.metadata.get("score", 0.0)
always resolves to 0.0. The threshold check score >= 0.5 always fails and the method always returns (None, 0.0).

Back in run(), this means:
doc_result, _ = self.get_relevant_document(question)
if doc_result:   # always False
    ...
else:
    first_response = first_chain.invoke(question)  # bare LLM, no context ever used
The RAG path (if doc_result) has never executed. The vector store does nothing.
```

## Defect 2  run() passes the wrong key to the chain (currently masked by Defect 1)
If Defect 1 is fixed in isolation, run() will crash immediately. The chain is constructed expecting the key "question":
chain_input = {
    "context": self.retriever | format_docs,
    "question": RunnablePassthrough()    # expects "question"
}
But the RAG branch invokes it with "query":
if doc_result:
    first_response = first_chain.invoke({
        "query": question,               # wrong key
        "context": doc_result.page_content
    })
RunnablePassthrough() receives nothing for "question". The prompt template renders {question} as an empty string and the LLM generates a nonsensical response.

This has never surfaced because Defect 1 ensures doc_result is always None and the if doc_result: branch never executes.

### My proposed fix involves two small changes:
1. **Fetching the FAISS score:** Bypassing the LangChain retriever abstraction and calling `similarity_search_with_score` directly on the underlying vector store. This allows us to actually retrieve the distance score, convert it to a similarity metric, and properly evaluate the threshold.
2. **Fixing the chain invocation key:** Updating the dictionary passed to the chain from `"query"` to `"question"` so that it matches the `RunnablePassthrough` key the chain was built with. This prevents the prompt template from rendering an empty string once the document is successfully retrieved.

rahulkate173:
### Fix for Defect 1 (Core Retrieval Bug)
* **Renamed Variable:** Changed `self.retriever` to `self.vector_store` to store the raw FAISS index instead of the LangChain `VectorStoreRetriever` wrapper.
* **Updated `setup_vectorstore()`:** Removed the `.as_retriever()` call and now assigns the FAISS object directly to `self.vector_store`.
* **Rewrote `get_relevant_document()`:**
  * Added a `None` guard to handle cases where the vector store hasn't been initialized.
  * Now calls `self.vector_store.similarity_search_with_score(query, k=1)` directly to retrieve `(Document, float)` tuples (where the float is the L2 distance).
  * Converted the L2 distance to a similarity score using `1.0 / (1.0 + distance)`. This ensures the threshold comparison works as expected (higher score = more similar).

### Note Regarding Defect 2
The reported wrong-key bug (`"query"` vs `"question"`) appears to have been resolved in a prior refactor. The `run()` method no longer uses a LangChain chain with `RunnablePassthrough`. Instead, it already formats the prompt directly via `self.prompt_template.format(question=..., context=...)` and calls `self.provider.generate()`. No further changes were required for this defect.

Closes #130


Noaman-Akhtar:
i went through both PR 160 and PR 161, and they look like two different solutions for the same current retrieval issue. am i missing some difference between them, or is PR 161 meant to replace PR 160?

rahulkate173:
Yes pr 161 is meant to replace pr 160

Noaman-Akhtar:
Then you should close pr 160 , so it doesn't create confusion

rahulkate173:
Done 

Pr 160 closed

PR #164, opened 2026-08-27
Use subprocess.run instead of os.system in webhook handler

### LINKED ISSUE #135: Critical Security Vulnerability: Command Injection in Webhook Handler
I found a potential command injection vulnerability in app/routes/webhook.py (around lines 106–143).

The webhook handler currently uses os.system() with string interpolation to execute git commands:

git_fetch_command = f"cd '{REPO_PATH_LOCALLY}' && {GIT_PATH} fetch origin main"
fetch_result = os.system(git_fetch_command)

**Why this is dangerous:**

os.system() executes commands through the shell
Shell command construction with formatted strings can lead to command injection
If environment variables such as REPO_PATH_LOCALLY or GIT_PATH are manipulated, arbitrary commands could potentially be executed on the server

Impact:

Arbitrary command execution
Possible server compromise
Violates secure coding best practices (OWASP)

**Suggested Fix:**
Replace os.system() with subprocess.run() using argument lists instead of shell strings.

Example:

import subprocess

subprocess.run(
    [GIT_PATH, "fetch", "origin", "main"],
    cwd=REPO_PATH_LOCALLY,
    check=True,
    capture_output=True,
    timeout=30
)

This avoids shell interpretation and is significantly safer.

References:

OWASP Command Injection
Python subprocess security best practices

sumitjhadev:
Fixes #135.

Replaces os.system() shell-string execution with parameterized
subprocess.run calls so webhook-triggered commands can't be
manipulated via shell metacharacters. See the commit message for
the reasoning behind the signature-parsing change and test coverage.

### DIFF
--- a/app/routes/webhook.py
+++ b/app/routes/webhook.py
@@ -20,6 +20,7 @@ from fastapi.responses import JSONResponse
 import logging
 import hmac, hashlib
 import os
+import subprocess
 from dotenv import load_dotenv
 
 router = APIRouter(tags=["webhook"])
@@ -61,8 +62,11 @@ def verify_github_signature(body: bytes, signature: str) -> bool:
         return False
     
     try:
-        sha_name, signature = signature.split('=')
-        if sha_name != 'sha256':
+        parts = signature.split('=', 1)
+        if len(parts) != 2:
+            return False
+        sha_name, signature_digest = parts
+        if sha_name != 'sha256' or not signature_digest:
             return False
         
         mac = hmac.new(
@@ -70,7 +74,7 @@ def verify_github_signature(body: bytes, signature: str) -> bool:
             msg=body, 
             digestmod=hashlib.sha256
         )
-        return hmac.compare_digest(mac.hexdigest(), signature)
+        return hmac.compare_digest(mac.hexdigest(), signature_digest)
     except Exception as e:
         logger.error(f"Error verifying signature: {e}")
         return False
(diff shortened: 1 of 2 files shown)


app/routes/webhook.py
```diff
@@ -33,7 +34,7 @@
 # Load secrets from env
 WEBHOOK_SECRET = os.getenv('WEBHOOK_SECRET')
 REPO_PATH_LOCALLY = os.getenv('REPO_PATH_LOCALLY')
-GIT_PATH = os.getenv('GIT_PATH')
+GIT_PATH = os.getenv('GIT_PATH', 'git')
```
chimosky: `GIT_PATH` can't be `git` because it's not a path, I'd rather you check `which git`.

app/routes/webhook.py
```diff
@@ -57,11 +58,11 @@ def verify_github_signature(body: bytes, signature: str) -> bool:
     """Verify GitHub webhook signature"""
     if not _webhook_configured:
         return False
-    if not signature:
+    if not signature or '=' not in signature:
```
chimosky: What's the idea behind this?

app/routes/webhook.py
```diff
@@ -57,11 +58,11 @@ def verify_github_signature(body: bytes, signature: str) -> bool:
...
         return False
     
     try:
-        sha_name, signature = signature.split('=')
+        sha_name, signature_hash = signature.split('=', 1)
```
chimosky: Why the `maxsplit` arg?

app/routes/webhook.py
```diff
@@ -58,20 +58,23 @@ def verify_github_signature(body: bytes, signature: str) -> bool:
...
     if not _webhook_configured:
         return False
-    if not signature or '=' not in signature:
+    if not signature:
         return False
     
     try:
-        sha_name, signature_hash = signature.split('=', 1)
-        if sha_name != 'sha256':
+        parts = signature.split('=', 1)
```
chimosky, line 65: Why the `maxsplit` arg?

sumitjhadev: I used `maxsplit=1` so the signature is split only at the first `=` into the algorithm name and the digest.

This keeps the parsing limited to the expected `sha256=<digest>` structure and avoids splitting the remainder unnecessarily. 

If you prefer a stricter or simpler parsing approach here, I’m happy to change it.

chimosky: Did you confirm that multiple `=` existed before now? I'm wondering why that'll be the case.

sumitjhadev: No, legitimate GitHub signatures never contain multiple = (they are hexadecimal, not base64).

maxsplit=1 was just added defensively for malformed headers. Since the enclosing try...except already catches unpacking errors and safely returns False, signature.split('=') is completely safe.

I can revert to signature.split('=') if you prefer the simpler version!

chimosky: I have no issue with your fix, I Just wanted you to articulate why, it'll be great to include that in your commit message. See [making commits](https://github.com/sugarlabs/docs/blob/master/src/contributing.md#making-commits).

sumitjhadev: Thanks for the pointer! Squashed everything into one commit and added the reasoning for both changes directly in the message the subprocess.run switch and why maxsplit=1 is there. 

Should be clearer for anyone looking back at this later.

sumitjhadev: Thanks so much for pointing me to the making-commits guide, @chimosky it really helped clarify things. 
I've reworked the commit message, title, and description around it. 
Kindly review it.

sumitjhadev:
Hi @chimosky, hope you're having a great week!

Just wanted to gently bump this when you get a moment. I updated the commit message and structure based on your feedback. No rush at all thanks again for your guidance on this!

app/tests/test_webhook.py
```diff
@@ -0,0 +1,260 @@
...
+from app.routes.webhook import verify_github_signature, webhook
+
+
+class TestWebhook(unittest.TestCase):
+    """Test suite for the GitHub Webhook endpoint security and functionality."""
+
+    def setUp(self):
+        self.secret = "test-webhook-secret-key"
+        self.repo_path = "/home/sugar/sugar-ai"
+        self.git_path = "/usr/bin/git"
```
chimosky, line 36: These two should never be hardcoded, your assumption of the `repo_path` is wrong, on another device, the directory would probably not exist, and I'd prefer you using `which git` to get the git path, that's a better way to do it.

app/tests/test_webhook.py
```diff
@@ -0,0 +1,260 @@
...
+        payload = b'{"ref": "refs/heads/main"}'
+        signature = self._generate_signature(payload)
+        self.assertTrue(verify_github_signature(payload, signature))
+
+    def test_verify_github_signature_invalid(self):
+        """Test that invalid signatures are rejected."""
+        payload = b'{"ref": "refs/heads/main"}'
+        self.assertFalse(verify_github_signature(payload, "sha256=invalid_hash_value"))
+        self.assertFalse(verify_github_signature(payload, "sha1=invalid_algorithm"))
+        self.assertFalse(verify_github_signature(payload, "invalid_format_without_equal"))
```
chimosky, line 58: A better thing to do here would be to use actual values, than just text explaining them.

app/tests/test_webhook.py
```diff
@@ -0,0 +1,260 @@
...
+        mock_request_missing = MagicMock()
+        mock_request_missing.body = AsyncMock(return_value=payload)
+        mock_request_missing.headers = {}
+        with self.assertRaises(HTTPException):
+            asyncio.run(webhook(mock_request_missing))
+
+        # 2. Malformed header
+        mock_request_malformed = MagicMock()
+        mock_request_malformed.body = AsyncMock(return_value=payload)
+        mock_request_malformed.headers = {"X-Hub-Signature-256": "malformed_no_equals"}
```
chimosky, line 111: Another place where a value would be preferred.

app/tests/test_webhook.py
```diff
@@ -0,0 +1,260 @@
...
+        with self.assertRaises(HTTPException):
+            asyncio.run(webhook(mock_request_malformed))
+
+        # 3. Invalid signature
+        mock_request_invalid = MagicMock()
+        mock_request_invalid.body = AsyncMock(return_value=payload)
+        mock_request_invalid.headers = {"X-Hub-Signature-256": "sha256=invalid"}
```
chimosky, line 118: Here too.

chimosky:
Reviewed, if you're using an LLM for your PRs, go through our [AI guide](https://github.com/sugarlabs/docs/blob/master/src/contributing.md#ai-guidelines-for-sugar-labs-contributions).

chimosky:
@Noaman-Akhtar could you go through all the tests here and be sure they're what they should, I've gone through some and you can see where I stopped from my comments.

Noaman-Akhtar:
> @Noaman-Akhtar could you go through all the tests here and be sure they're what they should, I've gone through some and you can see where I stopped from my comments.

sure, will go through them

PR #173, opened 2026-09-14
Feat/io Add API contracts and multimodal (image, audio) input

Noaman-Akhtar:
## Summary

Sugar-AI's endpoints had no defined contract. `/ask`, `/ask-llm` and
`/debug` took the question as an unvalidated query parameter, every
endpoint returned a differently shaped dict, errors were bare strings,
and message content was hardwired to `str` through every layer, which
also blocked any image or audio input.

This PR adds request and response contracts to every endpoint and builds
multimodal input on top of them. Message content can now be a list of
typed text, image and audio parts, validated at the edge and translated
into each provider's own wire format.

## What changes

**Output contracts**
- Every endpoint declares a Pydantic `response_model`. Field names are
  unchanged, so this formalizes existing shapes rather than altering them.
- All errors leave the API as `{"error": {"code", "message"}}` with
  stable codes (`unauthorized`, `validation_error`, `quota_exceeded`,
  `modality_not_supported`, and so on).

**Input contracts**
- `/ask`, `/ask-llm` and `/debug` accept a validated JSON body. The old
  query-parameter style still works as a deprecated fallback.
- The chat/prompted mode rules of `/ask-llm-prompted` move from handler
  `if` checks into a model validator.

**Multimodal input**
- `content` is a string or a list of parts:
  `{"type": "text" | "image" | "audio", ...}` with base64 data.
  Base64 is decoded during validation, mime types are allow-listed,
  images are capped at 5 MB and audio at 20 MB.
- Each provider declares what it accepts. Ollama is asked per model via
  `/api/show` at startup, since support there varies by model. Gemini
  defaults to text, image and audio. HuggingFace and OpenAI-compatible
  default to text, widenable through the new `AI_SUPPORTED_MODALITIES`
  setting, which always wins over detection. Unsupported media returns
  422 `modality_not_supported` naming what was sent and what the model
  accepts, instead of an upstream error. `/health` reports the active
  model's modalities.
- Providers translate parts to their own formats: OpenAI-compatible
  `image_url` and `input_audio`, Gemini `inline_data`, Ollama's `images`
  list. Ollama has no audio input, so audio is refused by the gate.
- `/ask-llm` and prompted mode take an optional `attachments` list.
  Chat mode takes parts directly inside messages.
- Audio is input only. Spoken replies need a different response shape
  and are left as future work.

## Compatibility

Deliberately additive. String content stays valid, query parameters
still work, response fields keep their names, and a text-only request
produces a byte-identical provider payload to before (covered by a
test). The one intentional change: `/ask-llm-prompted` mode violations
return 422 with the error envelope rather than 400 with a bare string.

## Testing

112 tests, all new (the repo had no tracked suite). They cover the error
envelope, every response shape, validation in both call styles, content
parts, modality gating and detection, and per-provider translation
against mocked transports asserting the exact payload built. An
end-to-end test sends a real HTTP request with an image through the
real route, schemas, `RAGAgent` and `GeminiProvider`, faking only the
network.

Verified live as well:
- Gemini (`gemini-3.1-flash-lite`): text, image and audio all answered
  correctly, including chat mode with an image.
- Ollama, vision model (`qwen3.5:0.8b`): image answered, audio refused
  with 422.
- Ollama, text-only model (`llama3.2:1b`): detection reports text only,
  image and audio both refused with 422.


vyagh:
I took a look at this, I don't think I'll have to do a lot of changes on my PR, if this is merged.
I will update my work for the new image format.
btw, I like https://github.com/Noaman-Akhtar/sugar-ai/pull/3 more, the error handling stuff from this can move there too.

